PR #1, opened 2016-02-12
Create a .editorconfig with 2 spaces indent for js

MetalMatze:


MetalMatze:
I've added HTML, CSS & Go.

The HTML & CSS styles are based on http://codeguide.co/ which I try to incorporate as much as I can.

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+insert_final_newline = true
+
+[*.{html,css,js}]
+charset = utf-8
+indent_style = space
+indent_size = 2
+
+[*.go]
+indent_style = tab
+indent_size = 8
```
tboerger: I would prefer 4 :)

MetalMatze: For `{html,css,js}`? If so, I really don't care. I just like that there's codeguide.co which one can point to if some code isn't properly formatted. Beyond just indentations.

tboerger: 2 for HTML is fine for me. I just prefer 4 for go

mikkeloscar: It doesn't really make sense to force an indent size when using tab.

tboerger: It does, but only visual within the editor

mikkeloscar: That is what i meant. It doesn't affect the code if you use 4 in your editor and I use 8 in mine, thus it shouldn't be forced in an `.editorconfig`.

PR #2, opened 2016-02-12
Update almost all npm deps and introduce webpack

MetalMatze:
Webpack is awesome for developing with react.
There's a `webpack -w` command which I've introduced as `make watch`.
It caches all modules and only recompiles the ones, that changed.
Building after changing a react component now takes ms instead of 5s.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,46 +2,46 @@
   "name": "drone-ui",
   "version": "1.0.0",
   "description": "",
-  "main": "index.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/drone/drone-ui.git"
   },
   "author": "",
-  "license": "ISC",
+  "license": "Apache",
   "bugs": {
     "url": "https://github.com/drone/drone-ui/issues"
   },
   "homepage": "https://github.com/drone/drone-ui",
   "dependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -2,46 +2,46 @@
...
   "version": "1.0.0",
   "description": "",
-  "main": "index.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/drone/drone-ui.git"
   },
   "author": "",
-  "license": "ISC",
+  "license": "Apache",
```
tboerger: I think that needs to be Apache-2.0

tboerger:
One tiny NITPICK, otherwise LGTM

package.json
MetalMatze: Thanks, that was just a quick fix I did without proper research. It's updated in this PR.

tboerger:
Still LGTM, but I don't have the permissions to merge :)

MetalMatze:
Added less support for webpack. I just hurts to run `make less` all the time. I want this to be automatic.

tboerger:
I dislike less in general :)

MetalMatze:
I like sass more! But without it's still a pain...

PR #3, opened 2016-02-14
Make --scheme optional in proxy server.

mikkeloscar:
This is just a simple addition to the proxyserver that makes it possible
to omit the `--scheme` flag and instead parse the scheme from the
`--host` parameter. The old behaviour still works.

(I also sneaked in a commit correcting the build instruction in the README)

### DIFF
--- a/server.go
+++ b/server.go
@@ -23,6 +23,12 @@ var (
 func main() {
 	flag.Parse()
 
+	hostScheme := strings.Split(*host, "://")
+	if len(hostScheme) == 2 {
+		*scheme = hostScheme[0]
+		*host = hostScheme[1]
+	}
+
 	// create the drone client and get the Drone user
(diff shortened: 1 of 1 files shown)


MetalMatze:
I've not used scheme. Did work perfectly. But this will probably clarify things.

mikkeloscar:
Yeah, you are right, the scheme flag can always be omitted if you are using http.

My motivation for adding this is that I can run the server against my drone instance which happens to be defined in `$DRONE_SERVER` and `$DRONE_TOKEN` because those are used with `drone-cli`.

So I run `go run server.go --host $DRONE_SERVER --token $DRONE_TOKEN` where `$DRONE_SERVER` includes `https://`

tboerger:
I would also prefer a single flag for scheme and host.

PR #10, opened 2016-06-25
Create a development environment with docker

MetalMatze:
Right now it's not that great to get started with the UI development.
I took the time to write a `docker-compose.yml` with `gogs:latest` and `drone:0.5`.

Sadly it's not working a 100% because the gogs container can't find the drone container to push webhooks... If anyone has an idea, please let me know!

Also to test the websocket implementation, one needs the current master from github.com/bradrydzewski/drone which isn't available as docker image, right @bradrydzewski ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /dist/
 /node_modules/
 .DS_Store
+.docker
--- a/README.md
+++ b/README.md
@@ -47,16 +47,16 @@ Just run `docker-compose up` in a separate terminal and it will start [Gogs](htt
 
 #### Gogs
 
-Now you have to setup Gogs. Gogs runs on `http://localhost:10081`. Just open that in your browser.
-All you need to change is to set the _Database Type_ to SQLite3 and _Application URL_ to `http://localhost:10081/`.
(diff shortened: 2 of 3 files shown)


bradrydzewski:
@MetalMatze perhaps try putting everything on the same network? you can use the ambassador pattern here or even host networking

```
services:
  drone:
    network_mode: host
  gogs:
    network_mode: host
```

or something like this (won't be exact)

```
services:
  network:
    image: busybox
    cmd: sleep 999d
  
  drone:
    network_mode: container:network

  gogs:
    network_mode: container:network
```

> to test the websocket implementation, one needs the current master from github.com/bradrydzewski/drone which isn't available as docker imag

true although this is a temporary problem. Once we have a working UI I can immediately merge my changes into master.

bradrydzewski:
also note that when I test locally I usually bypass github/ gogs all together. I have a payload (hook.json) stored locally and I manually trigger the hook:

```
TOKEN=....
curl -X POST -H "X-GitHub-Event: push" -H "Content-Type: application/json" -d @hook.json \
	"http://localhost:8080/hook?access_token=$TOKEN"
```

this ends up being much faster than having to constantly test the full workflow

PR #14, opened 2016-07-07
Refine webpack config and use webpack-dev-server

choffmeister:
Might not be final yet. Did also some overall reworking of the webpack configuration inspired by some knowledge I gained in my last two years working in this JS stack.

What do you think?


choffmeister:
Also there is still some design glitch in the topbar I don't know yet how I created it.

MetalMatze:
I'm currently not really sure if `drone:master` is the right branch?! 
I see that @bradrydzewski is really active on the `baobab` branch.

choffmeister:
Also recognized that. With the eslint fix commit there will be conflicts for sure. But we can leave that out for now. The first commit could work out for master branch. I tried my best to stay as compatible with the current setup as possible as build output etc goes.

bradrydzewski:
closing because we are using the `boabob` branch which has deviated significantly from master

choffmeister:
OK. But is there still interest in usage of webpack-dev-server? It really speeds up development cycles.

MetalMatze:
For me it looks interesting. But why did you create a gulpfile? I think including the webpack-dev-server and doing everything through npm scripts could be cool. Does the webpack-dev-server also have a proxy to the actual drone API?

bradrydzewski:
@choffmeister I'm primarily a Go develoepr and am not exremely knowledge when it comes to frontend tools and frameworks. I don't know what `webpack-dev-server` is. It would be beneficial to indicate what it does, and why we should include it.

also agree with @MetalMatze that I don't want to introduce gulp

choffmeister:
OK. I understand the gulp thing.

@bradrydzewski @MetalMatze The webpack-dev-server is basically a JavaScript http server, that loads the webpack configuration and starts a HTTP server. Developers can then access this devserver and will get the compile webpack assets as if they were build to disk. It also handles watching, recompiling and then signaling the browser to reload the page. Additionally it can act as proxy server to the drone API (including websockets). Basically what the `server.go` currently does.

Ofc webpack-dev-server can be used bare bone without gulp. I just copied over some stuff from my current project.

If you are interested in this (would lower the entry barrier for frontend devs, too) without gulp and just with npm run scripts, I would happy to provide another PR for this.

MetalMatze:
@choffmeister Totally. I do know a little about the webpack-dev-server. It the one also supporting hot reloading, right? I'd be really happy to see that in action!!!

choffmeister:
It does. Then I will rework my former PR.

PR #22, opened 2016-07-08
Add docker-compose.yml for develop

choffmeister:


bradrydzewski:
I actually would rather include a `docker-compose.yml` in the main drone documentation. This could really apply to drone development, in general, and this repository is really agnostic to the actual provider. I actually develop directly against beta.drone.io

choffmeister:
Also fine. Should I create a PR for drone itself then?

bradrydzewski:
cc @gtaylor do we have a `docker-compose.yml` included in the 0.5 documentation site? I think this would probably be helpful to those looking to contribute, to quickly compose a gogs instance for local testing. thoughts?

gtaylor:
I'm not sure if this is what you are asking, but we have this: https://github.com/drone/docs/blob/release/0.5/content/samples/installs.md

The compose example in there is github-based, but the repo could be sub-divided like I have with the Kubernetes repo: https://github.com/drone-demos/drone-on-kubernetes

bradrydzewski:
@choffmeister I appreciate the offer and appreciate everyone sharing their configurations, but I'm not sure it is something I want to commit to officially supporting or maintaining in the official drone repository. I think it is best that drone commit to providing a Docker image, and any third party tooling or configurations (mesos, kubernetes, compose, etc) are managed externally by community members. I'm happy to include links to such external repositories in the official drone documentation.

PR #23, opened 2016-07-08
Use webpack-dev-server

choffmeister:
Webpack-dev-server can replace the `server.go` and offers some nice features for frontend development:

* No need to leave JS world
* Watching and rebuilding (all in memory, so faster then writing stuff to disk)
* Hot reloading (change a file, will recompile and the browser will automatically reload)
* Acts as proxy to drone

But I have still some questions here.

1. Is the `static/` folder necessary or was it just an artifact form the go server
2. When going to the dashboard there is only the login link. Help is appreciated


package.json
```diff
@@ -26,15 +26,19 @@
...
+    "webpack-dev-server": "^1.14.1",
+    "yargs": "^4.7.1"
   },
   "scripts": {
-    "build": "NODE_ENV=production ./node_modules/.bin/webpack",
+    "start": "node ./webpack.devserver.js",
+    "build": "NODE_ENV=production webpack",
     "embed": "go-bindata-assetfs -o dist/app.go -ignore 'map|go' -pkg dist -nocompress=true dist/...",
-    "lint": "./node_modules/.bin/eslint src/",
-    "watch": "./node_modules/.bin/webpack -w"
```
choffmeister: The `./node_modules/.bin` folder is automatically appendend to the current path when doing a `npm run ...`

.gitignore
```diff
@@ -3,6 +3,7 @@
 /dist/*.ico
 /dist/*.map
 /dist/*.svg
+/dist/*.html
```
choffmeister: Is there a particular reason why not ignoring the whole `./dist/` folder?

choffmeister: Just saw that `.go` file are committed into the folder. Just ignore my question 😄

bradrydzewski: yep, and just to provide context, we are generated an `app.go` file that embeds our bundled css and js. This `app.go` file is imported into drone and compiled into the main binary.

index.html
```diff
@@ -7,15 +7,15 @@
   <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet"/>
   <link href="https://fonts.googleapis.com/css?family=Roboto+Mono" rel="stylesheet"/>
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet"/>
-  <link href="/static/app.css" rel="stylesheet"/>
-  <link href="/static/favicon.ico" rel="icon" type="image/x-icon"/>
+  <link href="/app.css" rel="stylesheet"/>
+  <link href="/favicon.ico" rel="icon" type="image/x-icon"/>
 </head>
```
bradrydzewski: could we keep these as `/static`? we need to load these from the `/static` subdirectory (or some prefix) in the drone server binary, so would be nice to use the same path in drone, and in the drone-ui server

bradrydzewski:
@choffmeister this is awesome work, thanks so much!  Added one minor comment and then let's get it merged!

index.html
choffmeister: Of course. Actually it would be good, if all static assets (that is everything but the index.html) were served from a certain subfolder. So I move everything (also the app.js and so on) into the `static/` folder

choffmeister: @bradrydzewski Do you have an idea, why on the dashboard there are no projects listed, even though the `window.STATE_FROM_SERVER` is filled properly?

choffmeister:
@bradrydzewski All assets are now in `dist/static/...` and the index.html is at `dist/index.html`

src/index.js
```diff
@@ -28,4 +28,9 @@ const RootedApp = root(tree, App);
 window.tree=tree;
 window.events=events;
 
-ReactDOM.render(<RootedApp />, document.querySelector('#app'));
+superagent.get('/api/user').then((res) => {
+  const user = JSON.parse(res.text);
+  window.STATE_FROM_SERVER = {user};
+
+  ReactDOM.render(<RootedApp />, document.querySelector('#app'));
+})
```
bradrydzewski: this should only be for the development version. Drone loads this automatically inside the index.html file on page load

choffmeister: True, wait a sec...

bradrydzewski: see https://github.com/drone/drone/blob/master/server/template/files/index.html for more details. We render the `index.html` server-side and inject the `user` and a `csrf-token` before serving to the user, avoiding additional round-trips / page flickering.

we just need a way to ensure we only load the user from the API in development mode, and not in production mode.

choffmeister:
Wrapped the part in a process.env.NODE_ENV if/else branch, see https://github.com/drone/drone-ui/pull/23/files#diff-1fdf421c05c1140f6d71444ea2b27638R31

webpack.config.js
```diff
@@ -38,27 +51,28 @@ module.exports = {
...
-      'process.env': {
-        NODE_ENV: '"production"'
-      }
-    }),
     new webpack.optimize.UglifyJsPlugin({
       compress: {
         warnings: false
+      },
+      output: {
+        semicolons: false
```
choffmeister: Newlines cost as much as semicolons in terms of filesize, but with newlines, stacktraces from production builds are a little more useful since not everything is put into a few lines.

src/components/layout/page.js
```diff
@@ -46,7 +46,7 @@ class Page extends React.Component {
               <Drawer>
                 <div className="brand">
                   <Link to="/">
-                      <Avatar className="logo" src="/static/drone.svg"/>
+                      <Avatar className="logo" src="/drone.svg"/>
```
bradrydzewski: do we need to adjust this link as well?

choffmeister: ofc

choffmeister: Sry, too much jumping between projects 😄

choffmeister:
Finally, should the `server.go` be removed with this PR? Or is it still needed?

bradrydzewski:
yes, can you confirm the node proxy is working including websockets? if so feel free to remove and update the PR

choffmeister:
drone-ui-http

bradrydzewski:
fyi I'm not seeing the live build stream proxy correctly. I can see the open websocket however no frames are being relayed

screen shot 2016-07-07 at 12 06 05 am

choffmeister:
OK, just took a look in the old `server.go` and I guess I found this issue: The token has to be appended as query parameter, not as header. That is why the connection cannot be established.

Will fix it asap.

bradrydzewski:
@choffmeister actually drone accepts both the `Authorization` header or query parameter. The reason we opt for query parameter from the browser is because the `WebSocket` connection in the browser does not provide the ability to set header variables.

I'm not sure about the node proxy server being used, but headers should work. This is the logic Drone is using to extract the token:

```Go
	var token = r.Header.Get("Authorization")

	// first we attempt to get the token from the
	// authorization header.
	if len(token) != 0 {
		token = r.Header.Get("Authorization")
		fmt.Sscanf(token, "Bearer %s", &token)
		return Parse(token, fn)
	}

	// then we attempt to get the token from the
	// access_token url query parameter
	token = r.FormValue("access_token")
	if len(token) != 0 {
		return Parse(token, fn)
	}

	// and finally we attempt to get the token from
	// the user session cookie
	cookie, err := r.Cookie("user_sess")
	if err != nil {
		return nil, err
	}
```

Thanks for taking a look!

bradrydzewski:
wondering if we need to use `ws` or `wss` as the proxy `target` server in `webpack.devserver.js` instead of http or https? also wondering if any of this is relevant: https://github.com/webpack/webpack-dev-server/issues/283

choffmeister:
Just tested the former `server.go` implementation (just had to remove the `http.StripPrefix("/static/",` part to make it work with the new folder structure) and there websocket does also not work.

screen shot 2016-07-11 at 07 34 07

screen shot 2016-07-11 at 07 34 17

choffmeister:
@bradrydzewski Maybe websockets isn't working at all in my docker-compose setup... What is a valid websocket URL that I should be able to access on my drone server?

bradrydzewski:
if you visit http://beta.drone.io in the browser you can see drone connect to the `/ws/feed` websocket endpoint

screen shot 2016-07-10 at 10 38 20 pm

bradrydzewski:
@choffmeister I invited you to the drone org which will give you access to beta.drone.io, if you want to use that for testing. Feel free to proxy directly against that instance if you want.

choffmeister:
@bradrydzewski Nice thanks. Could not login into beta.drone.io thought. Get an infinite relocate loop between drone and github. Anyway, I was right and my websocket was not working at all. After updating drone:0.5 docker image it is now working. I made a PR to fix the proxying, but cannot fully check if this really is working now, because I have to leave for work now. Will look into it later again.

PR #31, opened 2016-07-09
Serve all assets and prefix them with hashes

choffmeister:
This PR incorporates all external assets and serves them directly. Additionally all assets are prefixed with hashes.

This allows to set long term cache headers for the whole `static/` folder (`drone.svg` and `favicon.ico` are yet the only exceptions from that rule).

Why serving all assets directly and not from CDN? I suggest that many will run their CI in corporate environments. IMHO having the plus in privacy by not needing to request content from third parties outweights the small chance, to already have a certain file somewhere in cache. And even if not, since all assets have hashes and so can be long term cached, this is at most a concern for the very first visit of the UI.

### DIFF
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -4,11 +4,13 @@
   <meta charset="utf-8"/>
   <meta content="width=device-width, initial-scale=1" name="viewport"/>
   <meta content="ie=edge" http-equiv="x-ua-compatible"/>
+  <link href="/static/favicon.ico" rel="icon" type="image/x-icon"/>
   <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet"/>
   <link href="https://fonts.googleapis.com/css?family=Roboto+Mono" rel="stylesheet"/>
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet"/>
-  <link href="/static/app.css" rel="stylesheet"/>
-  <link href="/static/favicon.ico" rel="icon" type="image/x-icon"/>
+  <% for (var i = 0; i < htmlWebpackPlugin.files.css.length; i++) { %>
+  <link href="<%= htmlWebpackPlugin.files.css[i] %>" rel="stylesheet"/>
+  <% } %>
 </head>
 <body>
 <div id="app"></div>
@@ -16,6 +18,8 @@
   window.STATE_FROM_SERVER = {}
 </script>
 <script src="https://code.getmdl.io/1.1.3/material.min.js"></script>
-<script src="/static/app.js"></script>
+<% for (var i = 0; i < htmlWebpackPlugin.files.js.length; i++) { %>
+<script src="<%= htmlWebpackPlugin.files.js[i] %>"></script>
+<% } %>
 </body>
 </html>
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -5,9 +5,6 @@
   <meta content="width=device-width, initial-scale=1" name="viewport"/>
   <meta content="ie=edge" http-equiv="x-ua-compatible"/>
   <link href="/static/favicon.ico" rel="icon" type="image/x-icon"/>
-  <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet"/>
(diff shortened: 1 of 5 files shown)


choffmeister:
One more hint: Instead of using the material icon font as a whole (which is several ten kilobytes in size) we can also use the SVG sprites directly with [svg-react-loader](https://github.com/jhamlet/svg-react-loader). With this loader we can pick the individual icons and compile them directly into the JS code with ease. Since we are only using a few icons of all the icons available, this will reduce data transferred for the icons basically to nothing. But this might be up for another PR later on.

bradrydzewski:
@choffmeister using the hashed values is not currently an option. The `index.html` page in this project is not used by the drone server. Drone server uses a Go template that is rendered server-side and it isn't really practical at this time to update the server-side template every time the files in this repository change.

I'm super appreciative of all these suggestions and the improvements you're providing. I do recommend we discuss these sort of changes in advance in our gitter channel (see https://gitter.im/drone/drone-dev) for more efficient collaboration. Thanks!

bradrydzewski:
I moved the CDN suggestion to its own issue. See issue #32

tboerger:
I would suggest to integrate the cdn files but drop the hash prefix.

choffmeister:
Wouldn't it be possible to read the index.html from the generated app.go file? And then the server could just atring replace the STATE_FROM_SERVER into the read index.html.

This way we could use the hashed filenames and further more there would be no need at all to ever change something in drone just because there was the need for changes in the index.html.

tboerger:
That should be possible but that still increases the changes to the bindata a lot

choffmeister:
OK, but as far as I understand right now, every time something in the index.html changes, one has to also go to the drone code and change stuff there, too, right? So doing this change once (the reading of the index.html from app.go) would free up time in the long run and ensure, that drone/drone-ui are compatible in this regard as long as the STATE_FROM_SERVER stuff has not been stripped from the index.html.

choffmeister:
@bradrydzewski I would love to have commit access here. I am really interested in this project. But ofc I understand, that I yet have to prove, that I will not destroy stuff on a regular basis :)

bradrydzewski:
@choffmeister I'm open to using the `index.html` file from this project if we can find a way to generate the file with the require Go template syntax for injecting the csrf token and STATE_FROM_SERVER. Check out https://github.com/drone/drone/blob/master/server/template/files/index.html

choffmeister:
@bradrydzewski I updated my branch. The index html in development now looks so:

```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8"/>
  <meta content="width=device-width, initial-scale=1" name="viewport"/>
  <meta content="ie=edge" http-equiv="x-ua-compatible"/>
  
  <link href="/static/favicon.ico" rel="icon" type="image/x-icon"/>
  
  <link href="/static/app-2486b24078092b4afb06.css" rel="stylesheet"/>
  
</head>
<body>
<div id="app"></div>

<script>
  window.STATE_FROM_SERVER = {};
</script>


<script src="/static/app-2486b24078092b4afb06.js"></script>

</body>
</html>
```

whereas with `NODE_ENV=production` it looks so:

```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8"/>
  <meta content="width=device-width, initial-scale=1" name="viewport"/>
  <meta content="ie=edge" http-equiv="x-ua-compatible"/>
  
  {{ if .csrf }}<meta name="csrf-token" content="{{ .csrf }}" />{{ end }}
  
  <link href="/static/favicon.ico" rel="icon" type="image/x-icon"/>
  
  <link href="/static/app-77d1d281de5a5b730aff.css" rel="stylesheet"/>
  
</head>
<body>
<div id="app"></div>

<script>
  window.STATE_FROM_SERVER = {{ . | json }};
</script>


<script src="/static/app-77d1d281de5a5b730aff.js"></script>

</body>
</html>
```

So it matches https://github.com/drone/drone/blob/master/server/template/files/index.html and could be picked up by drone (whereas the template in drone could be dropped).

If this is OK, please reopen the PR (if this is even possible) or tell me to create a new one. Could then be merged when you have time to adapt drone project to this.

bradrydzewski:
it isn't letting me re-open the pull request so we'll need to re-create. thanks for the change!

PR #33, opened 2016-07-10
Serve all assets and postfix them with hashes

choffmeister:
Elaborated version of #31 /cc @bradrydzewski


bradrydzewski:
thanks! please allow a few days for this to get merged, since this will require server-side changes, and I have some higher priority items I need to complete first

choffmeister:
No problem. Nice that it will be merged eventually.

Viele Grüße
Christian Hoffmeister

bradrydzewski:
I wanted to provide a quick update here. I think the best way to enable this is to place the `login.html` template in this repository as well. Then we can add a `Template()` function to the `app.go` file to return a Go `template.Template` object with all compiled templates.

This will ensure all UI code is encapsulated in two functions `AssetFS` to get the assets and `Template` to get the templates. No UI code will be stored in drone server at that point.

tboerger:
Needs a rebase.

choffmeister:
Rebased to master.

vtolstov:
@choffmeister do you have any free time to fix merge conflict?

choffmeister:
@vtolstov Have rebased it on master. But have no time to check if this really works (only know, that `npm run build` works fine) right now.

vtolstov:
@tboerger @bradrydzewski can somebody test this? Or how can i test (i don't have nodejs)

choffmeister:
The development server seems to work. I ran `npm start -- --scheme http --host beta.drone.io:80 --token eyJh...` and (except for the repo list on the left), everything is fine:

screen shot 2017-02-23 at 09 24 52

choffmeister:
But I cannot test if the complete Go build works. I have not worked with go before.

tboerger:
LGTM

vtolstov:
I'm test this pr and have one issue  dist_gen.go have 27M and i think that it very big.
Mostly this is svg fonts. Does we really need it?

tboerger:
`dist_gen.go` includes all assets, so not much we can change here.

vtolstov:
@tboerger may be fonts can be served via cdn and if cdn dies fonts not matter, but js files serve from drone?

vtolstov:
@tboerger with all assets inside go file i can't compile drone on vps with 2Gb memory i think this is bad.

tboerger:
I personally dislike CDN parts especially for a software that can be hosted on an intranet.

vtolstov:
@tboerger i'm too , but 20Mb assets only for fonts ???

vtolstov:
@tboerger may be we can cleanup svg and some other extensions? keeping only needed for current stable browsers ?

vtolstov:
http://caniuse.com/#search=woff

vtolstov:
so we can have only woff or ttf fonts inside assets

tboerger:
Feel free to provide a POC with an improved webpack config.

vtolstov:
@tboerger sorry =(, may be in the future, i don't know webpack

choffmeister:
I can change the PR on the weekend to only carry `.woff` and `.ttf`

vtolstov:
@tboerger why not only ttf? as i see woff does not supported by android 4.3 but ttf does.

choffmeister:
@vtolstov But woff2 should still be included I think. It is way smaller then woff and we should rather let clients profit from smaller downloads than only focus on an as small as possible server image.

vtolstov:
@choffmeister i don't think so. In case of intranet - you don't worry about sizes. In case of global site - use cdn and smaller fonts? @tboerger what do you think?

choffmeister:
We are talking about 600kb for the .woff2 files in total in the server image, but a saving of around 33% in download size for the client (for every single client).

vtolstov:
@choffmeister ok, so woff2 + ttf ?

choffmeister:
woff2 is not supported by IE 11 (don't know if that matters), but we could just go for that and live with a fallback to a system font in that case.

choffmeister:
One more notice: IMO this web font-face stuff is outdated anyway. If you don't have a fashion homepage or something else that heavily relies on style and art, then the font is really not that important. Many pages (like for example GitHub) have moved away from downloaded fonts and just go with a feasible set of fonts, so that on every system there is at least one of these fonts already present. This does not hurt usability but saves a lot of download size when opening up a web page/web app.

For example GitHub uses `font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";`

vtolstov:
@choffmeister may be we need something like this? So in this case static asset have only js files and logo?

choffmeister:
Yes. So if no one insists on Roboto as font, this would be what I would suggest to go for.

jhasse:
> woff2 is not supported by IE 11 (don't know if that matters)

Since the UI already has quite a few bugs in any non-Blink browser (e. g. http://caniuse.com/#search=dialog), I doubt that pixel-perfect rendering in IE 11 matters at all ;)

PR #36, opened 2016-07-11
fix websocket proxying for dev

choffmeister:
Still not sure if this really fixes the problem, but I have to leave now. At least the 101 protocol update is now working.

### DIFF
--- a/webpack.devserver.js
+++ b/webpack.devserver.js
@@ -7,11 +7,13 @@ const webpack = require('webpack');
 const webpackConfig = require('./webpack.config.js');
 const WebpackDevServer = require('webpack-dev-server');
 
+const proxy = require('http-proxy').createProxyServer();
 const port = 9000;
 const drone = {
   scheme: argv.scheme,
   host: argv.host,
   server: `${argv.scheme}://${argv.host}`,
+  wsServer: `${argv.scheme === 'https' ? 'wss' : 'ws'}://${argv.host}`,
   token: argv.token
 };
 
@@ -32,16 +34,6 @@ const server = new WebpackDevServer(compiler, {
       },
       xfwd: true,
       changeOrigin: true
-    },
-    // proxy to drone websockets
-    '/ws/*': {
-      target: drone.server,
-      headers: {
-        'Authorization': `Bearer ${drone.token}`
-      },
-      xfwd: true,
-      ws: true,
-      changeOrigin: true
     }
   },
   // reduce the console noise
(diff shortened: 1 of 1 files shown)


choffmeister:
Looks good 👍 

drone-ws-proxy

bradrydzewski:
are you using `"http-proxy": "^1.14.0"`? It definitely looks like we're getting closer here although I'm seeing some errors. Any ideas? Wondering if there is a different between localhost proxy vs domain name beta.drone.io or something?

screen shot 2016-07-11 at 12 19 18 am

bradrydzewski:
ok I think I might know the issue. I'm not sure it is picking up the auth credentials. It successfully connects the server and opens the feed websocket, which does not require authentication.

```
time="2016-07-11T07:51:11Z" level=debug msg="Successfully upgraded websocket" 
time="2016-07-11T07:51:11Z" level=info ip="::1" latency=16.191197ms method=GET path="/api/user/feed" status=200 time="2016-07-11T07:51:11Z" user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.106 Safari/537.36" 
```

It then fails with a 401 unauthorized which happens when you try to access either a repository that does not exist or is private and requires authentication. My guess is that something is happening with this proxy request causing it to be sent unauthorized.

```
time="2016-07-11T08:00:41Z" level=info ip=50.137.170.110 latency=2.972481ms method=GET path="/ws/logs/drone/drone-test-go/253/2" status=401 time="2016-07-11T08:00:41Z" user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.106 Safari/537.36" 
```

choffmeister:
When I remove the token from the WS proxing it still works on my local setup. So in your and in my case there seem to be no auth token appendend (for my setup is looks no auth is needed?).

Sorry that this takes some many fixes of fixes. Will hopefully finally fix it tonight.

bradrydzewski:
@choffmeister no auth is required for the `/feed` or log streaming endpoints since you can view public repositories and public builds without having to authenticate. Auth is required for the `/feed` endpoint to send events for private repos, and for the log streaming endpoint to stream logs for a private repository.

I just pushed a fix that somewhat seems to work, which uses the header for websockets instead of the query parameter. So we seem to be very close here.

I do see the webpack-dev-server crash shortly after streaming the logs, and then clicking around a bit. Not sure the exact cause.

```

webpack: bundle is now VALID.
/Users/bradrydzewski/code/src/github.com/drone/drone-ui/node_modules/http-proxy/lib/http-proxy/index.js:119
    throw err;
    ^

Error: read ECONNRESET
    at exports._errnoException (util.js:873:11)
    at TCP.onread (net.js:557:26)
```

choffmeister:
@bradrydzewski OK, then the header stuff would have worked all the time. Thought this was the issue, but it was when the whole websocket stuff did not work in my setup. You can then also set

```
target: drone.wsServer,
```

since the proxy seem to only pickup the host anyway and ignores the path behind.

Yeah, I also see the server crashing. It might be because the websocket connection is resetted. `http-node` somehow thinks it is a good idea to completely crash the server by default, if a proxy connection fails. To make the server stay alive one has to register an `error` handler to ignore the error. Would know how to do that with `http-node` directly, but with webpackdevserver if have to find this out.

PR #37, opened 2016-07-11
fix webpack-dev-server crashes

choffmeister:
@bradrydzewski This should fix the crashing and hopefully the WDS is finally working as expected 🙏


choffmeister:
The websocket connection reset happens when the build finishes and so the logs stream is closed. This killed the proxy because of missing error handler.

Issue #15, opened 2016-07-08
JavaScript improvement ideas


choffmeister:
I studied the code a little bit and have some point that could/should be changed IMHO:

1. Use [react-virtualized](https://github.com/bvaughn/react-virtualized) for better performance (see https://github.com/drone/drone-ui/blob/master/src/pages/build/log.js)
1. Use higher level components to ensure data loading instead of `componentDidMount` (https://github.com/drone/drone-ui/blob/master/src/pages/build/content.js)
2. Use `Immutable.Record` instead of normalizer to ensure certain fields in data objects. This also comes with some nice additional features like custom domain specific actions on data objects.
2. Use a promise based request library that also handles response parsing (like JSON.stringify when content type is `application/json`) automatically. A nice choice is [axios](https://github.com/mzabriskie/axios) here
3. Replace all the `let`by `const`
4. Use library `classnames` to construct class name list, like `const className = classnames('status', {'success': item.success, 'failure': item.failure})` (see https://github.com/drone/drone-ui/blob/master/src/components/status.js)
5. Use `export {Header} from './header'` syntax (see https://github.com/drone/drone-ui/blob/master/src/pages/user_profile/index.js)
6. Name files after their default export. Hence `export default class Header` should be in pascal case `Header.js` 
7. Configure react-router with pure JSON instead of JSX (see https://github.com/drone/drone-ui/blob/master/src/routes.js)
8. Sort imports do disambiguate the choice (there is also a eslint rule to check that)

Sorry for this long list...

bradrydzewski:
I like all these ideas. The one piece I see un-necessary is the use of `Immutable.Record` because we are now using `baobab` and no longer using `immutable.js`. Baobab supports immutability by default.

> Note that the tree, being a persistent data structure, will shift the references of the objects it stores in order to enable immutable comparisons between one version of the state and another (this is especially useful when using strategies as such as React's pure rendering).

source https://github.com/Yomguithereal/baobab#updates

bradrydzewski:
I'm going to close this since I generally like to avoid larger catch-all issues. But let's definitely implement these suggestions. We can open separate issues if necessary, or simply just send individual PRs

choffmeister:
Fine. I will pick up task after task in the coming days/weeks as I find time.

bradrydzewski:
cool, thanks again for everything!

PR #41, opened 2016-07-11
Fix eslint errors

choffmeister:
Such commits are annoying I know, but better start with it earlier than later 😄 

See #15


choffmeister:
We (if) this is merged, then I would go and refactor some code to more functional. But before I would like to see some enforced linting.

bradrydzewski:
this is great, thanks!

PR #42, opened 2016-07-11
enforce import sorting with eslint

choffmeister:
See #15

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -16,10 +16,10 @@ rules:
   no-alert: 0
   no-console: 0
   quotes: [2, single]
+  react/jsx-no-undef: 2
   react/jsx-uses-react: 1
   react/jsx-uses-vars: 1
   semi: [2, always]
-  # TODO enable
-  # sort-imports: [2, {ignoreCase: true, ignoreMemberSort: true, memberSyntaxSortOrder: [all, single, multiple, none]}]
+  sort-imports: [2, {ignoreCase: true, ignoreMemberSort: true, memberSyntaxSortOrder: [all, single, multiple, none]}]
 globals:
   process: true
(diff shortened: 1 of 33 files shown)


choffmeister:
Also adds rule `react/jsx-no-undef` which throws a lint error, if trying to use a component in JSX, that is not defined. So one error source less.

bradrydzewski:
hey when you get a chance can you check any style changes? the custom style to ensure the left feed was 300px and the header colors and other custom styles (switches) look to have regressed

screen shot 2016-07-11 at 4 09 35 pm

bradrydzewski:
ok looks like it was some weird ordering issue. I added the following temporary fix, for now. I realize it isn't ideal, but I needed to unblock until we find a better way:

```diff
+/*eslint-disable */
+import './index.less';

import {events} from './actions/events';
import React from 'react';
import ReactDOM from 'react-dom';
import {root} from 'baobab-react/higher-order';
import {routes} from './routes';
import superagent from 'superagent';
import {tree} from './actions/tree';
import {Router, browserHistory} from 'react-router';

-import './index.less';
+/*eslint-enable */
```

choffmeister:
The import sorting rules if fine normally. The problem is the following:

For LESS the concatenation order matters. So in this case the `index.less` must be imported before the `routes`, because the `routes` again import other stuff that imports other LESS files. And so reordering here changes the output of the CSS styles in the end. This is really not ideal.

We should compose LESS with imports in LESS. And only import the main LESS file once in the `App.js`. This separates JS imports from LESS imports, which are really different in terms of order invariance.

bradrydzewski:
> We should compose LESS with imports in LESS. And only import the main LESS file once in the App.js. This separates JS imports from LESS imports, which are really different in terms of order invariance.

yep makes sense

PR #44, opened 2016-07-12
Add Dockerfile.

ogavrisevs:
Simple docker file on `node:argon` base image (node 4.4.7 version). Tested against drone `v0.5`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM node:argon
+
+# Create app directory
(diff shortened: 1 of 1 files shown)


bradrydzewski:
the UI gets embedded in the drone server and is not intended to run standalone, with the exception of local development. I appreciate the contribution and encourage you to use this file if you find it beneficial, but don't feel it is something we need to provide in the core repository.

PR #53, opened 2016-09-09
Provide different color for icon of pending jobs

MorrisJobke:
This then looks like this:

bildschirmfoto 2016-09-09 um 16 48 46

Before both icons for pending and running had the same color.

### DIFF
--- a/src/components/status.less
+++ b/src/components/status.less
@@ -38,10 +38,16 @@
   &.pending, &.running {
     color: #FFA000;
     border-color: #FFA000;
-    &:before {
-      content: "refresh";
-      animation: spinner 1.2s ease infinite;
-    }
+  }
+
+  &.pending:before {
+    content: "alarm";
+    animation: wrench 2.5s ease infinite;
+  }
+
+  &.running:before {
+    content: "refresh";
+    animation: spinner 1.2s ease infinite;
   }
 
   &.unknown {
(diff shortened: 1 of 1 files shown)


MorrisJobke:
@choffmeister @bradrydzewski Could you give any feedback on this? 😃

tboerger:
IMHO it should also be a different icon if you already start to change the color.

MorrisJobke:
Updated and it now looks like this:

bildschirmfoto 2016-09-26 um 13 17 13

The arrows are still spinning, while the clock is not animated at all.

@tboerger Is this okay? :)

tboerger:
LGTM 
@bradrydzewski?

bradrydzewski:
love the usage of the alarm clock. My recommendation would be to use the `faa-shake animated` [1] so that users don't think the page is frozen. I'd also recommend we keep the color yellow, since some people might interpret grey as disabled. I think having a different icon should be enough to differentiate. Thanks for the PR!

[1] http://l-lin.github.io/font-awesome-animation/

MorrisJobke:
@bradrydzewski Done. I only imported the keyframes for the shake animation instead of pulling in everything.

MorrisJobke:
@bradrydzewski @tboerger Anything else I need to address?

bradrydzewski:
sorry for the delay, looks good to me!

bradrydzewski:
note that these changes will automatically get included in drone the next time we run a build and publish to dockerhub

MorrisJobke:
Thanks :+1:

PR #55, opened 2016-09-24
Adds a message when user has not added any repositories yet

### LINKED ISSUE #51: Sidebar showing "loading..." if feed is empty
The sidebar in the dashboard shows a "Loading..." message if the feed is empty (length == 0).
The message is wrong. Is definitely not loading. I think it should show a more relevant message. Like suggesting to enable some projects with possibly a link to the relevant page.

gillesdemey:
screen shot 2016-09-24 at 20 50 26

fixes #51

### DIFF
--- a/src/actions/events.js
+++ b/src/actions/events.js
@@ -36,6 +36,8 @@ let token = function() {
 }();
 
 events.once(GET_FEED, function() {
+  tree.set(['pages', 'repo', 'loading'], true);
+
   Request.get('/api/user/feed?latest=true')
     .end((err, response) => {
       if (err != null) {
@@ -47,6 +49,7 @@ events.once(GET_FEED, function() {
         return (b.started_at || b.created_at || -1) - (a.started_at || a.created_at || -1);
       });
       tree.set('feed', feed);
+      tree.set(['pages', 'repo', 'loading'], false);
     });
 });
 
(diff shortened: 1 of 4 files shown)


tboerger:
This phrase is not entirely correct, it's also getting displayed if repos are activated but don't have a build yet. Otherwise lgtm

gillesdemey:
> This phrase is not entirely correct

Sure, just let me know what you'd like it to say :)

> it's also getting displayed if repos are activated but don't have a build yet

Wasn't able to reproduce this by adding a repo without a build, are you sure the length of `feed` would still be 0?

tboerger:
Yes I'm sure it will stay 0, but maybe brad is fine with it like it is.

gillesdemey:
I've noticed that it's briefly shown while data is being fetched from the API, I'm assuming that's what you meant.

Definitely not a great experience, it would probably make sense to introduce a `loading` property to the state of the sidebar.

Is this something we should attach to this PR or do we introduce it later?

bradrydzewski:
there is a loading field here that you might be able to use:
https://github.com/drone/drone-ui/blob/master/src/actions/tree.js#L67

bradrydzewski:
also we don't really use the blue button anywhere in the application, so it doesn't really fit the existing style. Let's instead mirror the same style below (gray text with dark link)

screen shot 2016-09-24 at 3 18 38 pm

gillesdemey:
I've addressed some of the concerns, PTAL :)

jadengore:
Any update on this?

bradrydzewski:
#88 was merged and fixes this. Thanks!

PR #57, opened 2016-10-01
Makes line numbers and time indicator non-selectable, fixes #54

### LINKED ISSUE #54: Copy logs to clipboard without line numbers
Due to line numbers.

gillesdemey:
As an added bonus it aligns line-numbers to the right (up to 9.999, which I think is a sane number)

### DIFF
--- a/src/components/term_row.js
+++ b/src/components/term_row.js
@@ -24,9 +24,9 @@ class TermRow extends React.Component {
 
     return (
      <div className="term-row">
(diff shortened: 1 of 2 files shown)


tboerger:
What about naming it term-num, term-line and term-time?

gillesdemey:
Sure, I have no strong preferences

bradrydzewski:
I ended up patching this locally. change is in master

PR #58, opened 2016-10-02
adding login button to the top right corner of the UI when logged out

dvmorris:


tboerger:
Do you have a screenshot?

dvmorris:
No, I don't actually know how to run this locally.

tboerger:
It's described in the readme how to launch the ui, you just need to point it to some drone server

dvmorris:
Oh OK, I will try that today and upload a screenshot.

dvmorris:
So, I got it running locally, but it seems that everything is hard-coded to always stay logged in. And there is no route defined for the logout button, and it's currently generating a javascript error in the console.

jhasse:
When running the UI via `npm start` one has to provide the auth token and this does result in always being logged in :/

Issue #47, opened 2016-07-13
build history item should include the timestamp


bradrydzewski:
Right now the build list does not include any sort of date / timestamp. This was omitted in the prototype because I wasn't sure what the layout would look like, but this certainly needs to be resolved before the 0.5 release.

marcinc:
I think this can be closed now.

PR #62, opened 2016-10-14
Adds time information for each build history item

marcinc:
This PR addresses #47.

### DIFF
--- a/src/components/build_meta.js
+++ b/src/components/build_meta.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import TimeAgo from 'react-timeago';
 
 import './build_meta.less';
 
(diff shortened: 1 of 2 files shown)


src/components/build_meta.js
```diff
@@ -12,6 +13,7 @@ class BuildMeta extends React.Component {
         <em>{build.author}</em>
         <span>pushed to</span>
         <em>{build.branch}</em>
+        <Timestamp time={build.created_at || build.enqueued_at} format="full" className="time"/>
```
bradrydzewski: we've been using `react-timeago` to display the build date in the user interface. Example at https://github.com/drone/drone-ui/blob/master/src/components/build_panel.js#L39:L51

We might want to do the same here for consistency

marcinc: Thought it'd be a good idea to show more detailed time information rather than just `X days ago`. But it also makes sense to keep time representation consistent across the board. Will change this to use `react-timeago`.

bradrydzewski:
can you please attach a screenshot so we can review the UI visuals as well? Thanks!

marcinc:
@bradrydzewski Screenshot attached. 
screen shot 2016-10-25 at 10 46 04

Issue #63, opened 2016-10-18
build history filtering


marcinc:
Currently there is nothing in UI which enables the user to filter the build history. 
I'd like to propose that filtering by basic build item properties is added to the UI so that it's easier to find relevant items. Initially filtering by `branch`, `author`, `event`, `tag`, `status`, `deploy_to`  should be sufficient but could be extended to include any build item attribute.

Some general top level ideas:
- Filtering criteria will be provided via tags e.g. `status:success`
- It should be possible to chain tags to build more complex filtering criteria, for example: `author:foo` `status:failure` ...
- It should be possible to add multiple filter components of the same attribute e.g. `tag:v1` `tag:v2`
- It should be easy to add and remove individual components from the filter criteria.
- All filter criteria components will be applied with logical AND operator, i.e. all conditions need to be met for the build history item to show up in the list. 
- Filter tag auto-completion (NTH)

tboerger:
Sounds interesting and somebody already talked about that on gitter

marcinc:
@tboerger yes, @vaijab mentioned that on gitter as we're looking into adding that. Am currently working on a prototype of that enhancement as it's something we'd like to use.

marcinc:
I've opened #64 with initial working version of build history filtering. Would be great to get some feedback on that. We've started using it internally and it's been quite useful. One extra addition to original proposal was the ability to pass filter in the query string so that it can be shared e.g.: 
`?filter=status:success;event:deployment`

bradrydzewski:
merged #64

strk:
@bradrydzewski was this support removed ? I tried this w/out success:
https://cloud.drone.io/postgis/postgis?filter=status%3Asuccess

PR #64, opened 2016-10-25
Adds repo build history filtering by base attributes

marcinc:
This (WIP) PR is related to enhancement proposal raised in https://github.com/drone/drone-ui/issues/63. Would be good to get some feedback on this.

Screencast:
filter

### DIFF
--- a/src/actions/events.js
+++ b/src/actions/events.js
@@ -1,3 +1,4 @@
+import * as FilterUtils from '../utils/filter';
 import Emmett from 'emmett';
 import Request from 'superagent';
 import {tree} from './tree';
@@ -22,6 +23,9 @@ export const GET_BUILD_LOGS = 'GET_BUILD_LOGS';
 export const DEL_BUILD_LOGS = 'DEL_BUILD_LOGS';
 export const FILTER = 'FILTER';
 export const FILTER_CLEAR = 'FILTER_CLEAR';
+export const BUILD_FILTER = 'BUILD_FILTER';
+export const BUILD_FILTER_CLEAR = 'BUILD_FILTER_CLEAR';
+export const BUILD_FILTER_SUGGESTIONS_CLEAR = 'BUILD_FILTER_SUGGESTIONS_CLEAR';
 export const GET_TOKEN = 'GET_TOKEN';
 export const SHOW_TOKEN = 'SHOW_TOKEN';
 export const HIDE_TOKEN = 'HIDE_TOKEN';
@@ -186,6 +190,11 @@ events.on(GET_BUILD_LIST, function(event) {
       builds.map(function(build) {
         tree.set(['builds', owner, name, build.number], build);
       });
+
+      // populate filter suggestions for fetched build list
+      FilterUtils.setSuggestions(builds);
+      // sets filtered_builds facet to dynamically select builds matching filter criteria
+      FilterUtils.filteredBuilds(owner, name);
     });
 });
 
@@ -403,6 +412,23 @@ events.on(FILTER, function(event) {
 
 events.on(FILTER_CLEAR, function() {
   tree.unset(['pages', 'repo', 'filter']);
+});
+
+events.on(BUILD_FILTER, function(event) {
+  const value = event.data;
+  if (value === '') {
+    tree.unset(['pages', 'repo', 'build_filter']);
+  } else {
+    tree.set(['pages', 'repo', 'build_filter'], value);
+  }
+});
+
+events.on(BUILD_FILTER_CLEAR, function() {
+  tree.unset(['pages', 'repo', 'build_filter']);
+});
+
+events.on(BUILD_FILTER_SUGGESTIONS_CLEAR, function() {
+  tree.unset(['pages', 'repo', 'suggestions']);
 });
 
 events.on(CLEAR_TOAST, function() {
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -21,8 +21,8 @@ export const tree = new Baobab({
   //
   repos: {},
 
-  // the build branch of the tree maintains an index of builds nodes organized
-  // by repository owner, name, and build number. For example:
+  // the build branch of the tree maintains an unfiltered index of builds nodes 
+  // organized by repository owner, name, and build number. For example:
   //
   // {
   //   "octocat": {
@@ -34,6 +34,11 @@ export const tree = new Baobab({
   // }
   //
   builds: {},
+
+  // the filtered_builds branch of the tree maintains a filtered index of builds nodes
+  // organized by repository owner, name and build number. Only nodes which meet filter
+  // criteria are included.
+  filtered_builds: {},
 
   // the feed branch of the tree maintains an activity feed array, consisting
   // of the latest activity per repository.
--- a/src/pages/repository/content.js
+++ b/src/pages/repository/content.js
@@ -3,20 +3,39 @@ import BuildCard from '../../components/build_card';
 import {Link} from 'react-router';
 import PageContent from '../../components/layout/content';
 import React from 'react';
-import {events, GET_REPO, GET_BUILD_LIST} from '../../actions/events';
+import Select from 'react-select';
+import {events, GET_REPO, GET_BUILD_LIST, BUILD_FILTER, BUILD_FILTER_CLEAR,
+        BUILD_FILTER_SUGGESTIONS_CLEAR} from '../../actions/events';
 
 import './index.less';
 
 class Content extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.onFilter = this.onFilter.bind(this);
+  }
+
   componentDidMount() {
     const {owner, name} = this.props.params;
     events.emit(GET_REPO, {owner, name});
     events.emit(GET_BUILD_LIST, {owner, name});
+    // if filter param is present in query string then use it
+    // e.g. ?filter=status:success;author:foo;event:deployment
+    const value = this.props.location.query.filter;
+    if(value) {
+      events.emit(BUILD_FILTER, value);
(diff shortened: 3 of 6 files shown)


bradrydzewski:
@marcinc looks really great!

@metalmatze or @choffmeister or @vaijab any change you would be interested in providing a code review, since the JS stuff isn't my expertise? Thanks!

@vaijab is this similar to what you guys were thinking?

vaijab:
> @vaijab is this similar to what you guys were thinking?

@bradrydzewski I am working together with @marcinc, same team.

src/pages/repository/content.js
```diff
@@ -3,20 +3,43 @@ import BuildCard from '../../components/build_card';
 import {Link} from 'react-router';
 import PageContent from '../../components/layout/content';
 import React from 'react';
-import {events, GET_REPO, GET_BUILD_LIST} from '../../actions/events';
+import Select from 'react-select';
+import {events, GET_REPO, GET_BUILD_LIST, BUILD_FILTER, BUILD_FILTER_CLEAR,
+        BUILD_FILTER_SUGGESTIONS, BUILD_FILTER_SUGGESTIONS_CLEAR} from '../../actions/events';
 
 import './index.less';
+import 'react-select/less/default.less';
```
tboerger: I would put that into index.less

src/pages/repository/content.js
```diff
@@ -28,9 +51,95 @@ class Content extends React.Component {
...
+    if (builds && (state == undefined || state.suggestions == undefined || state.suggestions.length == 0)) {
+      this.setFilterSuggestions();
+    }
+  }
+
+  setFilterSuggestions() {
+    const {owner, name} = this.props.params;
+    const {builds} = this.props;
+
+    var dict = {};
```
tboerger: Why a dict? Just to have unique values? Is that really performant?

src/pages/repository/content.js
```diff
@@ -28,9 +51,95 @@ class Content extends React.Component {
...
+    var func_body;
+    if(Object.keys(predicates).length > 0) {
+      var conditions = [];
+      Object.keys(predicates).forEach(k => {
+        conditions.push('(' + predicates[k].join(' || ') + ')');
+      });
+      func_body = 'return ' + conditions.join(' && ') + ';';
+    }
+
+    return new Function('o', func_body);
```
tboerger: I'm not sure if this is really nice :)

src/pages/repository/content.js
```diff
@@ -28,9 +51,95 @@ class Content extends React.Component {
...
+      }
+      if(item.deploy_to !== '') {
+        dict['deploy_to:' + item.deploy_to] = true;
+      }
+    });
+
+    var suggestions = Object.keys(dict).sort((a, b) => {return a > b;});
+    suggestions = suggestions.map(item => {return {label:item, value:item};});
+    events.emit(BUILD_FILTER_SUGGESTIONS, {owner, name, suggestions});
+  }
```
choffmeister: I would rewrite this mutating style code like so:

```js
const terms = Object.keys(build).map(k => builds[k]).reduce((dict, build) => {
  dict['author: ' + build.author] = true;
  return dict;
}, {});
```

Also please replace the `var`s with `const`s:

```js
const suggestions = Object.keys(terms)
  .sort((a, b) => a > b)
  .map(item => ({label:item, value:item}));
```

src/pages/repository/content.js
choffmeister: Yes this is really bad. You are basically constructing JS code strings and `eval` it for every single suggestion cycle...

marcinc: Agreed, I don't like it either. Will definitely rewrite that bit.

bradrydzewski:
I'm not sure if it makes sense, but one could also store the filtered list in the tree itself, using the computed data feature:
https://github.com/Yomguithereal/baobab#computed-data-or-monkey-business

The benefit here may be the fact that if the pane is re-rendered it won't need to re-calcuate / re-revaluate the list. Take this recommendation with a grain of salt though, since I definitely don't claim to be a JS expert.

marcinc:
@bradrydzewski Thanks! Will look into that. As a side note, I don't claim to be a JS expert as well :)

src/pages/repository/content.js
marcinc: Thanks. Will incorporate.

marcinc:
@bradrydzewski Filtered list is now dynamically calculated in the tree itself - it's much nicer this way. Filtering bits have been cleaned up too. I think this is now ready for a second look @choffmeister @tboerger.

tboerger:
This looks MUCH better now

src/actions/events.js
```diff
@@ -186,6 +191,91 @@ events.on(GET_BUILD_LIST, function(event) {
...
+          if(build.deploy_to !== '') {
+            dict['deploy_to:' + build.deploy_to] = true;
+          }
+          return dict;
+        }, {});
+
+        let suggestions = Object.keys(terms)
+          .sort((a, b) => a > b)
+          .map(item => ({label:item, value:item}));
+        events.emit(BUILD_FILTER_SUGGESTIONS, {owner, name, suggestions});
```
bradrydzewski: I'm not sure we should be emitting from here. I think we might need to continue to improve the approach. I will try to provide more feedback when I can spend some more time understanding the code

src/actions/events.js
```diff
@@ -186,6 +191,91 @@ events.on(GET_BUILD_LIST, function(event) {
...
+      })();
+
+      // parses filter and groups criteria by attribute
+      const filterCriteria = function(filter = '') {
+        const tags = ['tag', 'branch', 'author', 'status', 'event', 'deploy_to'];
+
+        return filter.split(';').reduce((criteria, item) => {
+          let tag = item.split(':');
```
bradrydzewski: I feel like we should be doing this in the component, and sending the tree the parsed value. This seems component-specific, as opposed to something the data management layer would concern itself with.

src/actions/events.js
```diff
@@ -405,6 +495,34 @@ events.on(FILTER_CLEAR, function() {
   tree.unset(['pages', 'repo', 'filter']);
 });
 
+events.on(BUILD_FILTER, function(event) {
+  const {owner, name, value} = event.data;
+  if (value === '') {
+    tree.unset(['pages', 'repo', owner, name, 'filter']);
+  } else {
+    tree.set(['pages', 'repo', owner, name, 'filter'], value);
+  }
```
bradrydzewski: There should only be a handful of possible children in `['pages']`, one for each page, which means we should have multiple entries for each repository:

```diff
-['pages', 'repo', owner, name, 'filter']
+['pages', 'repo', 'filter']
```

since only a single repository can be viewed at once, I'm not sure it makes sense to track the filter values per repository. Instead, I would suggest it is tracked in one place globally, and reset when the repository changes.

src/actions/tree.js
```diff
@@ -35,6 +35,11 @@ export const tree = new Baobab({
   //
   builds: {},
 
+  // the filtered_builds branch of the tree maintains a filtered index of builds nodes
+  // organized by repository owner, name and build number. Only nodes which meet filter
+  // criteria are included.
+  filtered_builds: {},
```
bradrydzewski: could the monkey function go here? I feel like a good amount of the logic can be removed from `events.js` and be placed here with a bit of refactoring and some minor design changes

src/actions/events.js
marcinc: This has been ported from the component so perhaps we could just set the value in the tree in here instead of emitting another event to do that? Not sure if triggering event from another event's handler is an anti-pattern here.

src/actions/events.js
marcinc: This could be a part of `filterPredicates` function really. Extracted it to keep functions size sane. This really just processes raw filter value stored in the tree to help building filter test functions.

src/actions/events.js
```diff
@@ -186,6 +191,91 @@ events.on(GET_BUILD_LIST, function(event) {
...
+            } else {
+              criteria[tag[0]] = [tag[1]];
+            }
+          }
+          return criteria;
+        }, {});
+      };
+
+      // returns filter test functions
+      const filterPredicates = function(filter) {
```
marcinc: Sure.

src/actions/events.js
marcinc: Scoping by `owner/name` gives ability to track filter values / suggestions per repository so in case one switches to another repo and filters by different attributes and then comes back to the previous one, the filters / filtered builds list will be maintained. 

This is different to filter used to select repo in the left hand sidebar: https://github.com/drone/drone-ui/blob/master/src/actions/events.js#L400

bradrydzewski: I personally would not want this to persist. If I leave the repo and come back I'd prefer the filters reset.

danielepolencic:
is this ready to be merged?

src/actions/events.js
marcinc: @bradrydzewski Updated. It resets filters when you leave repo page. If there is any more feedback on that please let me know. If all good I'll squash commits and change PR name to be no longer WIP.

marcinc:
Hi @bradrydzewski. Rebased and squashed. Appreciate the time is scarce but would be great to get some feedback and/or `lgtm` on this :)

bradrydzewski:
thanks!

bradrydzewski:
The bundled code was updated and is now available in `drone/drone:0.5`. I really appreciate your patience getting this merged. Thanks again!

ylecuyer:
@bradrydzewski why is the filter not present anymore in 0.8 ?

rushib1:
looks like the ui has been moved from react to vue and is still under progress.

PR #76, opened 2016-12-22
reset scroll position on navigate

### LINKED ISSUE #48: Scroll to top when switching pages
Sorry about the long title, I couldn't figure how to say it any more tersely.

If you are following a build and the build is long enough to scroll the page, when you navigate back to the list of the repositories past builds (on left), I expect to be shown the top of the list (most recent builds), however I get left scrolled down the page.

This requires me to scroll to the top to see by most recent build status.

cheslip:
this should fix scroll position when navigating between repos. Unfortunately a little hacky using `querySelector` but looks like it's a [limitation with `react-mdl`](https://github.com/react-mdl/react-mdl/issues/262)

fixes #48


jhasse:
Works for me :)

metalmatze:
If I have same behavior I mind that you try to fix here. I'm also for it, since I always need to scroll back to the top... 😄

PR #81, opened 2017-01-30
Keep aspect ratio of Drone logo

jhasse:
The Drone logo isn't an avatar (isn't square). I've increased the top margin so that this matches the avatar margin on the right.

Before:
before

After:
after


metalmatze:
That would be nice! I've seen this bug a few times. Didn't had the time to look into it.

metalmatze:
LGTM

PR #82, opened 2017-01-30
Center tab label in clickable area

jhasse:
Before:
before

After:
after


tboerger:
LGTM but the build needs to be restarted.

jhasse:
Sorry for the force pushes, I needed to change my email.

PR #86, opened 2017-02-01
Replace ansi_up with aha.js

jhasse:
See https://github.com/benbria/aha.js

This improves highlighting a lot, especially since ansi_up messes up with GCC's output and doesn't support bold or underlined text.

## Examples

Before:
before

After:
after

Before:
before2

After:
after2

I think that `dist/dist_gen.go` needs to be regenerated, didn't know how to do that.


vtolstov:
what about xterm.js ?

jhasse:
> what about xterm.js ?

AFAIK xterm.js emulates a "real" terminal which has width, etc. This enables stuff like progress bars, etc. As the build commands aren't run in the browser, but in the Docker container, this wouldn't work, since the UI only displays the log afterwards.

xterm.js also parses ANSI color codes, but using it only for that functionality would be a little bit overkill IMHO. It would also break stuff like smooth scrolling if I'm not mistaken.

vtolstov:
@jhasse yes, but it used in many places and have solid code base....

bradrydzewski:
> what about xterm.js ?

my understanding (and I could be completely wrong here) is that you need to define a fixed height for each terminal window. This would not work with drone, where logs automatically expand to show all content without scrolling, delegating scrolling to the window.

but if I'm wrong here, I'm open to looking at term.js or hterm.js as well.

keep in mind that we should also consider whether such libraries will allow us to collapse commands and link to line numbers, both of which are planned in future versions of drone.

package.json
```diff
@@ -4,7 +4,6 @@
   "description": "This project is a proposal for a react-based user interface for Drone.",
   "main": "webpack.config.js",
   "dependencies": {
-    "ansi_up": "^1.3.0",
     "baobab": "2.3.3",
```
bradrydzewski: add the following and run `npm install` locally. This will fetch the dependency from the node package manager, which is a central repository of packages:

```diff
"dependencies": {
 -    "ansi_up": "^1.3.0",
 +    "aha": "0.1.0",
      "baobab": "2.3.3",
```

src/utils/aha.js
```diff
@@ -0,0 +1,583 @@
+/*
```
bradrydzewski: remove this file

src/components/term_row.js
```diff
@@ -1,5 +1,5 @@
-import ansi_up from 'ansi_up';
 import React from 'react';
+import aha from '../utils/aha';
```
bradrydzewski: change to `import aha from 'aha';` which instructs node to fetch from the local dependencies downloaded from npm, declared in the package.json

src/components/term_row.less
```diff
@@ -37,38 +37,22 @@
...
+.term-row .bg-red    { background-color: red; }
+.term-row .bg-green  { background-color: green; }
+.term-row .bg-yellow { background-color: olive; }
+.term-row .bg-blue   { background-color: blue; }
+.term-row .bg-purple { background-color: purple; }
+.term-row .bg-cyan   { background-color: teal; }
+.term-row .bg-white  { background-color: gray; }
+.term-row .underline { text-decoration: underline; }
+.term-row .bold      { font-weight: bold; }
+.term-row .blink     { text-decoration: blink; }
```
bradrydzewski: does `term_row.js` support bright colors? not sure it is a huge problem if they do not, but if so, would need the styles added

jhasse: AFAIK bright and bold are the same for ANSI colors. For example `cmd.exe` on windows has no bold, only bright. On Linux bright is bold instead.

Drone 0.4 was using bold :)

jhasse:
I force pushed some changes based on your suggestions.

http://beta.drone.io/drone/drone-ui/101 still seems to build the old version :(

bradrydzewski:
Aside from the license (which I think is ok) it looks like `aha.js` might not support certain ansi bright variants. So we might be trading one library and a set of missing functionality, for another library with a set of missing functionality.

Note that we use `ansi_up.escape_for_html` to sanitize strings before html is injected into the document. This is critical from a security perspective. It is not clear to me at the moment that `aha.js` handles sanitization.

jhasse:
> Aside from the license (which I think should be ok, but is still not ideal) it looks like aha.js might not support certain ansi bright variants. So we might be trading one library and a set of missing functionality, for another library with a set of missing functionality.

See my other comment https://github.com/drone/drone-ui/pull/86#discussion_r98898726

For example http://www.perpetualpc.net/6429_colors.html#color_list lists the code 1 as "for brighter color", but most modern terminals (on Linux, macOS) will interpret it as bold. AFAIK bright is basically a fallback for fonts where bold isn't available (e. g. the Linux ctrl+tab+f2 tty).

> We also use ansi_up.escape_for_html to sanitize strings before html is injected into the document. This is critical from a security perspective. It is not clear to me at the moment that aha.js handles sanitization.

aha.js also does this :)
https://github.com/benbria/aha.js/blob/master/index.js#L543

bradrydzewski:
> For example http://www.perpetualpc.net/6429_colors.html#color_list lists the code 1 as "for brighter color", but most modern terminals (on Linux, macOS) will interpret it as bold. AFAIK bright is basically a fallback for fonts where bold isn't available (e. g. the Linux ctrl+tab+f2 tty).

got it, thanks for the explanation!

> aha.js also does this :) https://github.com/benbria/aha.js/blob/master/index.js#L543

great, thanks for finding this for me.

I think now we just need to figure out what we need to comply with the following clause of the mpl license:

> 3.1. Distribution of Source Form
>
> All distribution of Covered Software in Source Code Form, including any
Modifications that You create or to which You contribute, must be under
the terms of this License. You must inform recipients that the Source
Code Form of the Covered Software is governed by the terms of this
License, and how they can obtain a copy of this License. You may not
attempt to alter or restrict the recipients' rights in the Source Code
Form.
>
> 3.2. Distribution of Executable Form
> 
> If You distribute Covered Software in Executable Form then:
> 
> (a) such Covered Software must also be made available in Source Code
    Form, as described in Section 3.1, and You must inform recipients of
    the Executable Form how they can obtain a copy of such Source Code
    Form by reasonable means in a timely manner, at a charge no more
    than the cost of distribution to the recipient; and

jhasse:
The [build on beta.drone.io](http://beta.drone.io/drone/drone-ui/105) fails because of

```
> eslint src/
/drone/src/github.com/drone/drone-ui/src/components/term_row.js
  21:24  error  'Buffer' is not defined  no-undef
✖ 1 problem (1 error, 0 warnings)
```

in this line

```js
let html = aha(new Buffer((out || '')), opts);
```

Does someone know how to satisfy eslint in this case?

jhasse:
Okay got it: `import {Buffer} from 'buffer';`

The aha.js devs gave some feedback regarding the license: https://github.com/benbria/aha.js/issues/3
Looks good to me :)

bradrydzewski:
@jhasse unfortunately I think this will need to be discussed with an attorney. While I appreciate these individuals claiming it _should_ be fine to include the library in Drone, this is not enough for me to feel comfortable moving forward. These individuals (while certainly intelligent) are not attorneys and have not taken time to understand how the code will be bundled and distributed in Drone, nor have they advise how we must comply with the license (in terms of attribution, static vs dynamic linking, etc)

For example, the LGPL license could be interpreted to read that anything statically bundled inside Drone means that Drone becomes a derivative work and must therefore be licensed under LGPL [1]. Static linking of course becomes a bit murky when it comes to compiled JavaScript which is why an attorney would need to be consulted.

The library is also MPL 1.1 which does not address minified JavaScript code in the FAQ. This may be supported under the license but I do not have the legal expertise to read through and understand the differences between MPL 1.1 and MPL 2.0.

There is also the possibility that Drone will release an Enterprise Edition one day. I am not planning it at this time, but I'm still not ruling it out. It is unclear what implication a copyleft library would have, even a weak copyleft, should I decide to pursue this effort. Even if there isn't an issue bundling LGPL or MPL code, there are still compliance requirements such as attribution, links to source, etc.

So for these reasons we will not proceed until I have been advised by legal counsel on how to proceed.

[1] http://answers.google.com/answers/threadview/id/439136.html

jhasse:
Thank you very much for still considering this. I totally understand that you want to be on the safe side.

One last thing:
> The library is also MPL 1.1 which does not address minified JavaScript code in the FAQ. This may be supported under the license but I do not have the legal expertise to read through and understand the differences between MPL 1.1 and MPL 2.0.

The [MPL 1.1](https://www.mozilla.org/en-US/MPL/1.1/) states "[...] may be distributed only under the terms of this License or a future version of this License [...]" and that may be the reason for Mozilla only providing a 2.0 FAQ: MPL 1.1 projects can be distributed as 2.0.

PR #88, opened 2017-02-05
Show no repos enabled if results have returned with nothing

josmo:
if no repos are enabled it leaves Loading up.  This will show a diff notification while loading vs none being enabled.

### DIFF
--- a/src/actions/events.js
+++ b/src/actions/events.js
@@ -40,6 +40,7 @@ let token = function() {
 }();
 
 events.once(GET_FEED, function() {
+  tree.set(['pages', 'loading'], true);
   Request.get('/api/user/feed?latest=true')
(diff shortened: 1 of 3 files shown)


bradrydzewski:
We have a `loading` parameter in the tree that is intended for this purpose, but was never hooked up. If loading is false and the list is empty, it would be used to display an empty message.

josmo:
sounds good. I'm not exactly sure how the implementation should look like.  I'll take a look but if you have a quick suggestion I'll definitely appreciate it.

josmo:
@bradrydzewski updated to use the loading property. Let me know if that works for you.

josmo:
@bradrydzewski Just wanted to check in on this, if I used the correct property this time?

PR #90, opened 2017-02-09
improve line numbers rendering

bamarni:
It uses a min-width and right-aligns them so that every line which is a power of 10 doesn't cause the build output to be shifted.

Also, what about using native CSS counters? That could allow to simplify the backend not to calculate and send them in the future.

Before : 
drone-before

After :
dronenew

With this change the output seems to get shifted after the 100.000th line, is it acceptable or should it allow for more?


tboerger:
LGTM

jhasse:
> With this change the output seems to get shifted after the 100.000th line, is it acceptable or should it allow for more?

I'd even vote for less. Like 1,000 or 10,000. This would result in a little bit more horizontal space for the actual text and look better IMHO.

bamarni:
@jhasse : indeed that's the compromise, as we can see on the screenshot with my use-case it's already making line 13th to be a 2-liner.

10,000 sounds good to me, WDYT @bradrydzewski?

bradrydzewski:
> Also, what about using native CSS counters? That could allow to simplify the backend not to calculate and send them in the future.

using counter is not an option here because the ordering of the lines is not guaranteed. It is possible for the message broker to send messages out of order.

bamarni:
@bradrydzewski : Thanks for the info, does it means that currently output lines can be displayed in the wrong order but with correct line numbers?

In any case I've removed the css counter and reduced the min-width to 30px, which would display numbers until 9999 without a glitch.

bamarni:
@bradrydzewski : any news about this one?

Issue #79, opened 2017-01-18
Prompt user when deactivating a repo


jmccann:
We have had people "accidentally" deactivate repos.

For me this is usually due to a person being one of multiple Admins to a repo and not understanding Drone as well as others.  They want to "cleanup" their Drone UI and think untoggling a repo impacts only them and their UI view.

Would like a prompt to a user for confirmation when deactivating a repo.  With a message such as:

> Are you sure you want to deactivate <repo>?  This will prevent future builds from occurring as well as delete any web hooks, secrets and build logs for <repo>.

bradrydzewski:
@jmccann sorry for the delay here. The reason I haven't moved forward with this is that I've been kicking around this design change https://github.com/drone/drone/issues/2036

bradrydzewski:
ok, sorry for the delay on this. In master when you de-activate a repository it now toggle a `repo_active` flag in the database instead of deleting the repository 😄

tonglil:
@bradrydzewski is there a way to delete the repository via the UI still? Or must be done via CLI?

tboerger:
I thought the same because i wanted to entirely reset a project :)

tonglil:
^same use case

bradrydzewski:
We do have an endpoint that will delete the repository from the database. You could curl the endpoint as a temporary solution, in the meantime we should get it added to the CLI for our next release.

```
DELETE /api/repos/{owner}/{repo}?remove=true
```

PR #99, opened 2017-04-03
Add confirmation for deleting a repo

jmccann:
This is an attempt to address #79 

The thing I dislike the most is how I prevent the switch from "unswitching" if the user cancels the delete.  This is my first foray into React.  Feel free to give me some pointers.  😄


jhasse:
Just tried it: Works fine :)

Nit-pick: You could squash the two commits and add `, fix #79` to the commit message.

PR #106, opened 2017-05-02
remove external download material.min.js

binarytemple:
I've been working with drone in a docker-compose network. When I turn off my Wifi I found the drone-server was unusable, the reason is that the dropdown:

* Dashboard
* Account 
* Logout

Is dependent on material-ui, which is downloaded from a CDN. When offline it doesn't render anything so is unusable.

What I've done is to add 'material' as an NPM dependency (package.js) and configure webpack (webpack.config.js) to add it's contents to app.js. 

Let me know what you think.


metalmatze:
I think you installed the wrong dependency.
You should run `npm install material-design-lite --save` for the right. This is taken from https://getmdl.io/started/
Otherwise LGTM.

binarytemple:
Hi corrected the dependency as suggested, tested it locally using:

```
node webpack.devserver.js  --host=127.0.0.1:8000 --scheme=http --token=XXXXXXXXXXX
```

src/index.js
```diff
@@ -1,6 +1,8 @@
 /*eslint-disable */
 import './index.less';
 
+const Material = require('material-design-lite')
```
tboerger: Mixing `import` and `require`? :-1:

webpack.config.js
```diff
@@ -7,6 +7,7 @@ var webpack = require('webpack');
 module.exports = {
   entry: {
     app: [
+      'react-mdl/extra/material.min.js',
```
tboerger: Why? Just integrate it into the bundle.

Issue #115, opened 2017-05-18
Drone 0.6 do not display build duration


mavimo:
## Actual

Build duration is not visibile, eg:

<img width="280" alt="screen shot 2017-05-18 at 19 26 18" src="https://cloud.githubusercontent.com/assets/43941/26215260/f7fd2318-3bff-11e7-93c7-8e9bb4047400.png">

## Expected

I can see the build duration, eg:
<img width="291" alt="screen shot 2017-05-18 at 19 26 07" src="https://cloud.githubusercontent.com/assets/43941/26215242/e8a0b268-3bff-11e7-9156-8d47a6ebc6d9.png">

bradrydzewski:
closing as resolved / merged into 0.7

PR #116, opened 2017-05-18
refs #115: fix build duration

mavimo:
Fix build duration visibility (see issue)

cc/ @bradrydzewski

### DIFF
--- a/src/components/build_panel.js
+++ b/src/components/build_panel.js
@@ -56,8 +56,8 @@ class BuildPanel extends React.Component {
           <div>
             <div>
               <i className="material-icons">access_time</i>
-              {job.started_at ?
-                <TimeAgo date={(job.started_at || build.created_at) * 1000} /> :
+              {job.end_time ?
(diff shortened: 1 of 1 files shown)


bradrydzewski:
Thanks, this is definitely a bug, but I think we need to analyze this more. It looks like the timestamp and duration are both incorrect. See this screenshot from 0.5 when it was last working:

image

The top value should display the timestamp and bottom value should display the duration.

Right now it looks like the top value is always blank with a `--` and the bottom value displays the timestamp.

mavimo:
Opps, i supposed to display duration and timestamp, I'm going to fix it.

mavimo:
@bradrydzewski fixed

bradrydzewski:
This ended up being a bit more complex since the fix also needed to take matrix build status into account. I ended up patching locally and pushing.

Thanks for locating this issue though. I was able to build on your patch to get this fixed!

Issue #105, opened 2017-05-02
When adding secrets in the UI for a pem style variable it doesn't use it correctly for the plugin.


josmo:
Drone version:
.6

Steps to repro:
 1. go to the secrets section and add the contents of a pem file and save.
 2. run a build

Results:
The build will fail
ie. In the SFTP plugin it will result in ssh: no key found.

Need to have an option to format correctly or have it know out to.

bradrydzewski:
would a `<textarea>` solve this issue?

josmo:
I'll give it a try when I'm off the boat:)

PR #118, opened 2017-05-20
Updating of secrets now uses PATCH, also uses multiline text for upda…

josmo:
…ting of multiline secrets line pem values

@bradrydzewski This should fix the update of secrets since it looks like the endpoint changed.
I've also changed the textfields to multiline so it allows pem files and fixes https://github.com/drone/drone-ui/issues/105


josmo:
@bradrydzewski Just wanted to check on this to see if you were going to merge it in or needed anything?

bradrydzewski:
@josmo sorry this didn't end up making the 0.7 cut and I am in the process of porting the UI away from react-mdl which is now deprecated. Unfortunately the 0.8 UI is based on polymer (from an earlier prototype https://github.com/drone/drone-ui/tree/exp/polymer) which means these changes cannot be merged with the upcoming version.

The best bet is to re-issue the PR once I publish the 0.8 user interface, which will be in the new few days.

PR #119, opened 2017-05-21
fix rendering of matrix builds with parents (fork builds)

jashandeep-sohi:
`BuildPanel` component is not being passed the `repo` props on matrix builds. This breaks the rending of matrix builds that are "forked" (i.e. has a parent).

Passing the props downstream fixes this bug.

### DIFF
--- a/src/pages/build/matrix.js
+++ b/src/pages/build/matrix.js
@@ -19,7 +19,7 @@ export class Matrix extends React.Component {
 
     return (
       <PageContent fluid className="build">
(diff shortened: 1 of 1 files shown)


bradrydzewski:
THANKS!

This one has been annoying me and I haven't had time to hunt it down. I really appreciate you digging into the code and sending a patch 😄

bradrydzewski:
fyi I am going to tag a 0.7 version this week which will include this fix

PR #120, opened 2017-05-23
Webpack2

josmo:
@bradrydzewski I was upgrading another project to webpack 2 so I thought I try this one out at the same time.

Let me know if this works for you or if you'd like something changed :) I think I got everything that needs to be changed modified.


bradrydzewski:
the 0.8 version no longer uses webpack. So while I definitely appreciate the PR it is not something we'll need going forward. Thanks though! :)

metalmatze:
What's the next version going to use? :thinking:

metalmatze:
Nevermind, I just saw #126
Sorry for the noise.

PR #124, opened 2017-06-03
Allow output lines with long words to be broken

weargoggles:
When there's a long string of characters without spaces in the build log, it falls off the edge of the page. Setting min-width to 0, per [this article](https://hackernoon.com/11-things-i-learned-reading-the-flexbox-spec-5f0c799c776b), allows these long strings to be wrapped.

Before:
image

After:
image


bradrydzewski:
nice!!!

Issue #132, opened 2017-07-23
Log layout issues in Safari 10


bradrydzewski:
There seem to be issues with log layout in Safari 10. Unfortunately I don't have Safari 10/11 and am therefore unable to submit a patch for this issue. Help wanted.

Screenshot from Safari (broken)

![safari](https://files.gitter.im/drone/drone/XAG4/Screen-Shot-2017-07-22-at-10.19.22-AM.png)

Screenshot from Chrome (working)

![chrome](https://files.gitter.im/drone/drone/TDwN/Screen-Shot-2017-07-22-at-11.26.55-AM.png)

cc @jmccann

wass3r:
Submitted a fix https://github.com/drone/drone-ui/pull/133 - I've seen this type of behavior in IE before :(

PR #133, opened 2017-07-25
(css) Fix .term-row styling issue

wass3r:
Fixes https://github.com/drone/drone-ui/issues/132

So, this is *one* version of the fix. There is an alternative, which is arguably less desirable. The jist of the alternative solution is the following:

Change
```
<div class="term-row">
  <div></div>
  <div></div>
  <div></div>
</div>
```
to
```
<div class="term-row"><div></div><div></div><div></div></div>
```

This alternative solution also hints at the problem (at least in Safari, maybe others?). HTML code formatting is counted when the parent has ```white-space: nowrap``` in Safari (luckily Chrome seems to understand how to handle this as expected). New lines and multiple spaces are honored, both of which can and do exist in the HTML output inside ```.term-row``` elements and messes with the flex layout.

This change moves the formatting rule  from the parent container to target specifically the div that contains actual log output where formatting via white-space is useful.

Thanks @jmccann for helping with troubleshooting.

### DIFF
--- a/src/drone/drone-proc.html
+++ b/src/drone/drone-proc.html
@@ -46,8 +46,6 @@
 			color: #212121;
 			font-size: 12px;
 			line-height: 19px;
-			white-space: pre-wrap;
-			word-wrap: break-word;
 			font-family: Roboto Mono,monospace;
 			display: flex;
 			max-width:100%;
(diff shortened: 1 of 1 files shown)


bradrydzewski:
@jmccann @wass3r thanks so much, I really appreciate you taking the time to fix and submit a patch :)

PR #137, opened 2017-07-28
Add a link to Documentation

mrueg:
My javascript foo is weak, but this should add a page that provides a link to the Documentation in the user feed below if I parsed the existing code correctly, right?

### DIFF
--- a/src/drone/drone-app.html
+++ b/src/drone/drone-app.html
@@ -78,6 +78,10 @@
 			</app-route>
 
 			<app-route
+			        name='docs'
+				path='/account/docs'>
+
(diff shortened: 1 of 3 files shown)


metalmatze:
Not sure why this is closed, but it might be a good idea to do this. Prometheus for example has a link to their docs in the UI too.
This might also reduce the amount of people searching for the docs and finding a wrong version. We can always link to the appropriate version from within the UI. :tada:

mrueg:
I agree it could still be useful, but this change was targetting the polymer and not the react UI.

PR #146, opened 2017-08-17
Fix scrollToBottom on Firefox

cedk:


bradrydzewski:
reverted the codebase and this is no longer an issue

PR #147, opened 2017-08-21
Update drone-actions.html

gboddin:
This bit is causing issues on Firefox :
```
SyntaxError: expected expression, got ')'
```


bradrydzewski:
we moved this file to https://github.com/drone/drone-js which runs the code through babel and then packages for es5. So we should be all set :)

PR #154, opened 2017-09-09
Add props that's needed in "deployment" build

mavimo:
When there are deployment build `target` is missing


mavimo:
Without fix:

pre-fix

bradrydzewski:
thanks!

PR #157, opened 2017-09-10
Color consistency

mavimo:
Hi @bradrydzewski there are some color inconsistency so I try to uniformate styling across all UI.

1. Define a color palette, actually:
    - `#ffffff`: white
    - `#eceff1`: gray-light
    - `#bdbdbd`: gray
    - `#212121`: gray-dark
    - `#000000`: black
    - `#4dc89a`: green
    - `#fdb835`: yellow
    - `#fc4758`: red

1. Define colors as variable in `colors.less`
1. Include this file when needed
1. Replace colors with defined variables

During last step i find a lot of similar colors (eg `#222222` Vs `#212121`, ...) and individuate some inconsistency. I try to do my best to fix it maintaining the same visual impact, only really small changes are visibile (eg the vertical bar between left menu and body, flag in repo list, ..) but I think are improvement and not issue, let me know what's your opinion.

PS1: I'm not a designer, maybe difference in codebase was needed for some reason..
PS2: as extra value is extremely easy to customize tyle "per company" to have suuuuper nice CI server  ;)
PS2:  the last element is missing is logo personalization (doable by change in JS file)

### DIFF
--- a/src/screens/layout.less
+++ b/src/screens/layout.less
@@ -1,7 +1,7 @@
-@import "../shared/styles/colors.less";
+@import "~shared/styles/colors.less";
 
 .title {
-	border-bottom: 1px solid @gray-extralight;
+	border-bottom: 1px solid @gray-light;
 	display: flex;
 	box-sizing: border-box;
 	height: 60px;
@@ -46,7 +46,7 @@
   overflow:hidden;
 	overflow-y: auto;
   bottom:0px;
-  border-right:1px solid @gray-extralight;
+  border-right:1px solid @gray-light;
   box-sizing: border-box;
 }
 
--- a/src/screens/layout.less
+++ b/src/screens/layout.less
@@ -11,7 +11,7 @@
 	&> :first-child {
 		flex: 1;
 	}
-	padding:0px 20px;
+	padding: 0px 20px;
 
 	.avatar {
 		display: flex;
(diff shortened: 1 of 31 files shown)


mavimo:
PS: as extra value is extremely easy to customize tyle "per company" to have suuuuper nice CI server  ;)

PS2:  the last element is missing is logo personalization (doable by change JS file)

bradrydzewski:
Thanks! this is awesome work and very much appreciated.

src/screens/layout.less
```diff
@@ -45,7 +46,7 @@
   overflow:hidden;
 	overflow-y: auto;
   bottom:0px;
-  border-right:1px solid #CFD8DC;
```
bradrydzewski: This color is intentional, to draw contrast between the left and center panel, which can have unrelated content. Maybe we can create some sort of special variable for this? `@splitter-border-color: #CFD8DC` or something like that?

bradrydzewski:
one minor comment, but otherwise great and good to merge :)

mavimo:
@bradrydzewski we can use to `darken(@gray, 20%)`

Personally I prefer have the same color used on other separator (more consistent); difference is really limited -I note it only in refactoring switching browser between old/new screen- and I'm using drone every day from more then 1 year :)

Let me know your preference and I'll fix it ASAP.

bradrydzewski:
The color change for the separator was recommended by a UX designer friend of mine. I would like to consult with her before reverting back. I recommend we keep #CFD8DC as-is for now. I will discuss with her next week and then we can issue a follow-up pull request as needed.

Thanks :)

mavimo:
@bradrydzewski done

I do only a really small change, instead of define the color using HEX code i used darken on gray-light color, this allow to have consistency if someone try to change base colors; result is `#cfd6db` instead of `#cfd8dc`.

PR #160, opened 2017-09-10
Add JS linter

mavimo:
From readme you suggest to run `yarn run lint` but this feature is missing (I suppose after the last refactoring).

I start to add linting feature using standard tools (`eslint` and `eslint-plugin-react`).
Adding this check and using default config suggested from `eslint-plugin-react` (see https://raw.githubusercontent.com/yannickcr/eslint-plugin-react/master/.eslintrc ) there are a lot of changes that can be automatically fixed from eslint, but before that I want to know if there are some preference on styleguide.

I see this project use tabs (I suppose you're using the same editor for go project ;) ) but JS world prefer two spaces.. let me know if I can procede with standard rules or you prefer to change rules before lint the project.

cc/ @tboerger @bradrydzewski 

FYI after running linting to auto-fix most of the issue there are ~250 issues (most of them are unused variables, string too long, ...) that need to be manually fixed. I'll work on it if you agree on introduce this tool, but since the work to do is not small I prefer to be sure it wil accepted :)


bradrydzewski:
I have a local branch with eslint + prettier enabled. Let me push that and we can tweak the rules as needed. Thanks :)

mavimo:
@bradrydzewski as you can see eslint is a bit more aggressive :)

bradrydzewski:
yes :) I think we can enable most of those eslint rules. We just need to make sure we don't enable rules that conflict with prettier's auto formatting. I'll check to see if the prettier documentation provides a list of safe rules.

bradrydzewski:
>  the current codebase report me 181 errors using yarn lint, no error on your side

Yes, I get these errors as well

tboerger:
I think when you enable the react module you can drop most custom rules

mavimo:
WIP
 - fixed ~150 errors
 - missing `propTypes` in component is moved from error to warning (I think we can move it into a different PR)
 - added plugins to handle react, annotation and jest

.eslintrc.js
```diff
@@ -1,16 +1,32 @@
...
+		"plugin:react/recommended",
+		"prettier",
+		"prettier/react"
 	],
 	"plugins": [
+		"react",
+		"jest",
 		"prettier"
 	],
+    "parser": "babel-eslint",
```
tboerger: Wrong indentation :)

mavimo: Thx :)

PS: this is WIP, maybe you should not spend too much time in PR until finished :)

mavimo:
@tboerger @bradrydzewski JS lint pass on current codebase! 🎉 

Props are now warning (168 warning, I'll try to fix it on a different PR).

I tested UI functionality:
 - [x] Enable a new repo
 - [x] Disable an enabled repo
 - [x] Add a secret
 - [x] Remove an existing secret
 - [x] Add a new registry

bradrydzewski:
awesome work!  will get this merged today :)

PR #190, opened 2018-01-24
Use type="password" for secret

tizmagik:
This prevents certain browser extensions (e.g. [Grammarly](https://chrome.google.com/webstore/detail/grammarly-for-chrome/kbfnbcaeplbcioakkpcpgfkobkghlhen?hl=en)) from parsing/reading the secret value pasted into the `password` field.

### DIFF
--- a/src/screens/repo/screens/secrets/components/form.js
+++ b/src/screens/repo/screens/secrets/components/form.js
@@ -52,8 +52,8 @@ export class Form extends Component {
 					placeholder="Secret Name"
 					onChange={this._handleNameChange}
 				/>
-				<textarea
-					rows="1"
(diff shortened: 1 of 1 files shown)


bradrydzewski:
The textarea is important for multi-line secrets, such as ssh keys and jwt tokens, which are quite commonly used. We need a solution that solves for multi-line input.

tboerger:
Maybe a javascript lib that replaces the real value with `*` including an icon/button that can toggle the output.

tizmagik:
Ah, hadn’t considered the multi-line character issue, that’s a good point. Do we know for sure that newline characters would be swallowed with a password input type?

PR #194, opened 2018-02-13
Fix/disable access 4 user not loggedin

ycliuhw:
even read access should not allow for non authenticated users.

### DIFF
--- a/src/screens/drone.js
+++ b/src/screens/drone.js
@@ -1,19 +1,17 @@
+import { BrowserRouter, Route, Switch } from "react-router-dom";
+import { LoginError, LoginForm } from "screens/login/screens";
 import React, { Component } from "react";
-
-import { root } from "baobab-react/higher-order";
-import tree from "config/state";
-import client from "config/client";
-import { drone } from "config/client/inject";
-import { LoginForm, LoginError } from "screens/login/screens";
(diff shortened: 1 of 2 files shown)


tboerger:
If I see it correctly you always hide the builds if a user is not logged in? If so this is wrong behavior. Generally builds are visible for anonymous users, except for private repositories.

ycliuhw:
Hi @tboerger
The reason why I have this PR is I found the helm plugin logs all the global environment variables includes alll credentials(k8s cluster token for example). 
But if this is the expected behaviour for public repo, I am not sure we can hide credentials like *** when we stream the log to frontend page?

ycliuhw:
I think drone is really a well designed and modern replacement for Jenkins for containers. And I want to contribute to this project when I have time. But I am not sure if we have a roadmap for this ui repo already or not coz I saw there are a few different exp branch for vue etc. and also the master react branch doesn’t use redux etc data store.

tboerger:
It's up to the plugin to not printing secrets, drone doesn't know what is secure and what not

bradrydzewski:
I am going to close this pull request because drone is behaving as intended. I would recommend that you instead consider sending a patch to the helm plugin so that it does not print sensitive data to the console. Cheers

PR #191, opened 2018-02-01
add promote build to build menu

wavezhang:
Add a "Promote Build" menu to build drawer, when click it will toggle environments list display as follows. The deploy environment list is obtain from a modified droneServer api, see [this](https://github.com/drone/drone/pull/2332). Deploy parameters are not supported currently.
image

PR #195, opened 2018-02-20
add ability to page build list

### LINKED ISSUE #183: Infinite scroll of history of builds
The build history list only shows the last 50 builds.

Is there a way to get the next page of historical builds?

wass3r:
fixes #183 

depends on:
https://github.com/drone/drone-js/pull/3

a potential solution that adds simple "Show more builds" button at the bottom of the build list as discussed @ https://github.com/drone/drone/pull/2331


wass3r:
small preview of what this PR adds to the builds page:
paging

tonglil:
Thanks @wass3r!

tonglil:
PTAL @bradrydzewski.

bradrydzewski:
The only question I have is whether or not we should store the pagination details in the component state. Right now we store everything in the global store. I wonder (and I could be wrong here) what happens if react tries to re-use the component when you switch between repositories. The component state would not reset, in this case, and could result in the pagination not being reset between route changes.

wass3r:
@bradrydzewski State is per route/repo. If you navigate to a different repo you will see the first 50 unless you had paged that repo before. Let me know if you'd like it to work differently. I can see a case for either.

src/screens/repo/screens/builds/index.less
```diff
@@ -1,3 +1,24 @@
+@import "~shared/styles/colors";
+
 .root {
-	padding: 20px;
+  padding: 20px;
+}
+
+button {
+  background: @white;
+  border: 1px solid @gray-dark;
```
mavimo: I'll suggest to use `@gray-light` instead of `@gray-dary` to use the same convention we have on other boxed text

src/screens/repo/screens/builds/index.less
```diff
@@ -1,3 +1,24 @@
+@import "~shared/styles/colors";
+
 .root {
-	padding: 20px;
+  padding: 20px;
```
mavimo: Please use `yarn run lint` to lint LESS (eg: tabs instead of spaces)

tonglil:
Personally I would prefer to keep the pagination already loaded on a per repo basis.

src/screens/repo/screens/builds/index.less
wass3r: I think I borrowed the styling from some forms to try and keep things consistent, either from [here]( https://github.com/drone/drone-ui/blob/master/src/screens/repo/screens/registry/components/form.less#L39) or [here]( https://github.com/drone/drone-ui/blob/master/src/screens/repo/screens/secrets/components/form.less#L39).

I do see the action/log buttons have `@gray-light` or `@gray` but they contain SVGs (see [here](https://github.com/drone/drone-ui/blob/master/src/screens/repo/screens/build/logs/index.less#L30) and [here](https://github.com/drone/drone-ui/blob/master/src/screens/repo/screens/build/logs/index.less#L71)).

src/screens/repo/screens/builds/index.less
wass3r: I ran `lesshint` (I think that's what you meant). Didn't complain about tabs, but about double quotes. I did change to tabs for consistency though. Thanks for the pointer and taking a look.

src/screens/repo/screens/builds/index.less
mavimo: @wass3r actually we have `@gray-dark` border on button that indicate form submit (eg: secret settings for a repo) and `@gray-light` on boxed element like "follow log" button in build page.
IMHO this element have the same "meaning" that "follow build log" have, so I prefer light border, but I not stronger opinion about it. Thanks for investigate it 👍

bradrydzewski:
In terms of style, I have something specific in mind and will adjust accordingly once merged, so it is not something we need to focus on too heavily in this pull request.

In terms of state, I prefer to store state in the central store for consistency. With that being said, I'm not sure we need to store the pagination data in state or in the store. I think it can be derived.

We store the build data in the following structure:

```
["builds", "data", "octocat/hello-world", { 1: {...}, 2: {...}  }]
```

In the `fetchNextBuildPage` we could calculate the next page to fetch based on the number of builds in the current list. For example:

```
var page = Math.floor( builds.length / 50 ) + 1;
```

wass3r:
@bradrydzewski thanks for the feedback/suggestion! I did make some tweaks. To note, I'm passing the already created list (array) to the `fetchNextBuildPage` instead of recreating the list there, so it can be used in the page formula as you suggested. Let me know if this needs more tweaks.

jmccann:
@bradrydzewski Any updates on this?  😸

bradrydzewski:
@wass3r  @jmccann it looks like the package.lock file has a lot of updates. Is it possible to only update the drone-js dependency?

wass3r:
@bradrydzewski for sure. my fault. also, should the version for drone-js be incremented?

bradrydzewski:
yes, just bumped to drone-js@1.1.0

wass3r:
@bradrydzewski sorry for the delays, i think this does it?

mavimo:
Hi @wass3r @jesselang @jmccann @tonglil,
 since we have possibility to use drone with custom UI specifying the `DRONE_WWW` env that contains the folder with the UI we can integrate our UI changes very easily using a container with only UI exposed as volume. 

I create a PoC that include:

  - [#191](https://github.com/drone/drone-ui/pull/191): Add promote build to build menu
  - [#195](https://github.com/drone/drone-ui/pull/195): Add ability to page build list
  - [#200](https://github.com/drone/drone-ui/pull/200): Allow to set events when adding new secret

This can be tested using the following config (supposing to use `docker-compose` if you use kubernetes or some other orchestration tool please you should use a similar approach):

```yaml
version: '3'

services:
  drone-server:
    image: drone/drone:0.8.5
    environment:
      - DRONE_WWW=/drone-ui
    volumes:
      - drone-ui:/drone-ui

  drone-ui:
    image: mavimo/drone-ui:next
    volumes:
      - drone-ui:/drone-ui

volumes:
  drone-ui:
```

NB: the PoC I do is only intended useful for testing purpose, identify issue or improvement needed before PRs are included in mainstream, I do not suppose to do maintenance or have a fork that diverge from official UI, so if you chose to use it please report your feedback on issues on order to make it more usable :)

PR #200, opened 2018-04-04
Allow to set events when adding new secret

lafriks:
Default is left as it was before when adding from UI

Screenshot:
attels

### DIFF
--- a/src/screens/repo/screens/secrets/components/form.js
+++ b/src/screens/repo/screens/secrets/components/form.js
@@ -1,4 +1,12 @@
 import React, { Component } from "react";
+
+import {
+	EVENT_PUSH,
+	EVENT_TAG,
+	EVENT_PULL_REQUEST,
+	EVENT_DEPLOY,
+} from "shared/constants/events";
+
 import styles from "./form.less";
 
 export class Form extends Component {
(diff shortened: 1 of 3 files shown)


lafriks:
Yes but later as it needs new component for list of inputs

tboerger:
IMHO a text field with a comma or space separated list should be totally fine for the beginning.

lafriks:
textarea would be than probably be better with image per line

lafriks:
You want me to update that in this PR?

src/screens/repo/screens/secrets/components/form.js
```diff
@@ -59,6 +90,47 @@ export class Form extends Component {
 					placeholder="Secret Value"
 					onChange={this._handleValueChange}
 				/>
+				<section>
+					<h2>Events</h2>
+					<div>
+						<label>
+							<input
+								type="checkbox"
```
mavimo: Create a list of event / label than iterate over it instead of create list of markup.

mavimo:
Hi @lafriks @tboerger ,
 since we have possibility to use drone with custom UI specifying the `DRONE_WWW` env that contains the folder with the UI we can integrate our UI changes very easily using a container with only UI exposed as volume. 

I create a PoC that include:

  - [#191](https://github.com/drone/drone-ui/pull/191): Add promote build to build menu
  - [#195](https://github.com/drone/drone-ui/pull/195): Add ability to page build list
  - [#200](https://github.com/drone/drone-ui/pull/200): Allow to set events when adding new secret

This can be tested using the following config (supposing to use `docker-compose` if you use kubernetes or some other orchestration tool please you should use a similar approach):

```yaml
version: '3'

services:
  drone-server:
    image: drone/drone:0.8.5
    environment:
      - DRONE_WWW=/drone-ui
    volumes:
      - drone-ui:/drone-ui

  drone-ui:
    image: mavimo/drone-ui:next
    volumes:
      - drone-ui:/drone-ui

volumes:
  drone-ui:
```

NB: the PoC I do is only intended useful for testing purpose, identify issue or improvement needed before PRs are included in mainstream, I do not suppose to do maintenance or have a fork that diverge from official UI, so if you chose to use it please report your feedback on issues on order to make it more usable :)

PR #205, opened 2018-08-09
Set event pull_request when adding new secret

jimmykuo:
It lacks `PULL_REQUEST`

drone-secrets


tboerger:
IMHO this isn't required anymore because there should be already checkboxes for every event.

bradrydzewski:
for security reasons, this is not something we can enable by default.

> Secrets are not exposed to pull requests by default. Please be careful when exposing secrets to pull requests. If your repository is open source and accepts pull requests your secrets are not safe. A bad actor can submit a malicious pull request that exposes your secrets. The recommended approach to secure secrets is to enable gated builds.

PR #206, opened 2018-09-14
Don't forcibly lowercase case-sensitive secret names

greylurk:
Secret names when included in .drone.yml files are case sensitive. Lowercasing them in the UI leads to user confusion when the .yml file doesn't match the UI.

### DIFF
--- a/src/screens/repo/screens/secrets/components/list.less
+++ b/src/screens/repo/screens/secrets/components/list.less
@@ -18,7 +18,6 @@
 		flex: 1 1 auto;
 		font-size: 15px;
 		line-height: 32px;
-		text-transform: lowercase;
 	}
 
 	&> div:last-child {


bradrydzewski:
hey there, secrets are case-insensitive and are always passed to the plugin/container as uppercase environment variables. So this would probably not have the desired effect. This is something we'll have fixed in 0.9. You can track progress at https://github.com/drone/drone/issues/2070

greylurk:
That doesn't fit with my experience.  I was attempting to use a secret today and it was failing because I was using it in upper case in the drone.yaml, fitting with the pattern established earlier on the project I was working with, but when I looked in the UI, it was showing both the new and old secrets as lower case. When I viewed source in the UI, or used the CLI to examine the secrets, the old secrets were in upper-case, both in the yaml and in the secret store. However, the new secret name was in upper case in the yaml, and in lower case in the secret store.  

This resulted in the secret *not* being populated in my plugin.

PR #207, opened 2018-10-15
don't lowercase secret names in UI

gimlids:
Variable names in most programming languages I use are case sensitive. When Drone UI presents names of secrets in lowercase using a CSS lowercase transform, I become scared that Drone has actually lowercased my desired variable name (which would be very cruel indeed), or whether Drone is simply presenting my desired variable name in lowercase (less cruel, but still pretty cruel).

### DIFF
--- a/src/screens/repo/screens/registry/components/list.less
+++ b/src/screens/repo/screens/registry/components/list.less
@@ -18,7 +18,6 @@
 		flex: 1 1 auto;
 		font-size: 15px;
 		line-height: 32px;
-		text-transform: lowercase;
 	}
 
 	&> div:last-child {


gimlids:
I understand now that Drone subsequently uppercases secret names that are actually referenced in `.drone.yml`. I'm still a bit confused as to why the case changing happens at all. If some environment does not have case sensitive environment variable names, shouldn't that environment do any case changing?

```
$ env low=foo node --eval "console.log(process.env.LOW)"
undefined
$ env low=foo node --eval "console.log(process.env.low)"
foo
$
```

bradrydzewski:
The historical reason behind this behavior is because secrets are passed to the plugin and are case-sensitive. This means if you use the wrong case the plugin is not able to read the secret and the plugin fails. The case-sensitive nature of secrets resulted in a very high number of false-positive issue requests being created, so the problem was solved by forcing all secret variable names to uppercase.

In the 0.9 alpha version of Drone we have changed the syntax for how secrets are passed to plugins, and as a result, this workaround is no longer necessary.

before:

```
secrets:
- docker_username
- docker_password
```

after:

```
settings:
  repo: octocat/hello-world
  username:
    from_secret: username
  password:
    from_secret: password
```

I'm going to close this since we aren't planning any changes to 0.8 and instead are focused on 0.9, where this change would no longer be required.

PR #257, opened 2019-02-10
Display build.title for build title

sdbondi:
In Github (and perhaps others) PR builds the PR body/description is used as the title which makes for really long and not easily identifiable titles on the build list (especially with PR templates). This changes the build tiles to use `build.title || build.message`.  

*Problem:* 
screen shot 2019-02-10 at 10 33 57 etc etc ...

I'm not fully aware of the repercussions for this change, but this will work for PR and push builds which don't have the title attribute.

### DIFF
--- a/src/screens/repo/screens/build/components/details.js
+++ b/src/screens/repo/screens/build/components/details.js
@@ -14,7 +14,7 @@ export class Details extends Component {
 			<div className={styles.info}>
 				<StatusLabel status={build.status} />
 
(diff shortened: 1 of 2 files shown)


bradrydzewski:
this could also be somewhat related to https://github.com/drone/drone-ui/issues/260. @proAlexandr you good with this change?

bradrydzewski:
thanks, this has been resolved in a separate patch

PR #278, opened 2019-04-23
Show promotion target in build activity feed

### LINKED ISSUE #276: Show promotion target in build activity feed
We have a few environments we deploy to regularly, like `master`, `dev`, `staging`, and `prod`.  After we promote a build, drone does not show the target environment of the promotion in the build activity feed. As a result, it's hard to scan the build list to see which build corresponds to which target. The drone 0.8 ui showed the build target, and it seems missing from the v1 ui.

Concretely, right now the repo listing looks like:
image

It'd be helpful if it looked something like this instead:
image

sohan:
Fixes #276

### DIFF
--- a/src/components/RepoItemLabel.vue
+++ b/src/components/RepoItemLabel.vue
@@ -36,6 +36,9 @@ export default {
         if (event === "push" || event === "pull_request") {
           return { text: this.branch, href: this.hrefBranch };
         }
+        if (event === "promote" && this.promotionTarget) {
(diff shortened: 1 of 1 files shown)


bradrydzewski:
LGTM

sohan:
Hi @bradrydzewski anything else that needs to be done to merge this? Also it says only those with write access to the repo can merge pull requests -- would you just merge this branch or is there another process?

PR #279, opened 2019-04-23
When promoting a build, show the user who promoted the build

### LINKED ISSUE #277: When promoting a build, show the user who promoted the build instead of the author of the commit
If Sally checks in some code, and Bob promotes the build, the activity feed reads "Sally promoted `master`". This is a bit confusing, because Bob is the actor who did the promotion. We think the entry should read "Bob promoted `master`"

sohan:
Fixes #277 

When the build event `promote`:
* the text for the actor is `build.trigger`, instead of `build.author_login`
* the avatar is empty, since the backend doesn't pass an avatar that represents the value of `build.trigger`

Here's what it looks like in action:
image

Open questions:
* Is it ok to show no avatar instead of the avatar of the user who triggered the event? Should we just use a placeholder avatar instead?


src/components/RepoItem.vue
```diff
@@ -27,9 +27,9 @@
       </div>
 
       <div v-if="build" class="build">
-        <img :src="avatar" alt="avatar"/>
+        <img v-if="avatar" :src="avatar" alt="avatar"/>
```
bradrydzewski: I would prefer that we still display the avatar of the commit author. We can modify the Drone server to track deploy author and avatar in a future release, and then replace with the avatar if it makes sense.

sohan: Sounds good, reverted the changes to the avatar!

sohan:
Hi @bradrydzewski same questions -- anything else we need to do here? What's the best way to merge this?

bradrydzewski:
pending review from @lnikell

src/components/ReposPopup.vue
```diff
@@ -51,6 +53,9 @@ export default {
     }
   },
   methods: {
+    getBuildActor(build) {
```
proAlexandr: just `getBuildActor,` should be enough. Like here https://github.com/drone/drone-ui/blob/vue/src/components/AlertError.vue#L19

src/components/RepoItem.vue
```diff
@@ -29,7 +29,7 @@
       <div v-if="build" class="build">
         <img :src="avatar" alt="avatar"/>
         <div class="description">
-          <span>{{build.author_login}}</span>
+          <span>{{actor}}</span>
```
proAlexandr: It isn't clear to me why we should to create `getBuildActor` and pass its result as `actor` prop. I think it's simpler and clearer to write
```
{{build.event === "promote" ? build.trigger : build.author_login}}
```
right in the RepoItem.
I am sure that it's okay to place `if` right in the `RepoItem` because it's a common case, not a special logic, and, for example, we do so in the `RepoItemLabel`

sohan: My thinking was that this bit of code is used in a few components, so if we ever had to update the logic to get the actor, it'd be easier and less error-prone to just change this one function. Also it's probably a style preference, but imo the template code is simpler if it just reads `actor` since the prop name documents its contents. As someone reading the template, I can quickly visualize that the contents of the component are actor -> action -> link. Finally, to me `getBuildActor` felt similar to the `isBuildFinished` function in `buildHelper.js,` so I added it there.

This is not a super strong opinion and I'm happy to change it -- just wanted to share my thought process. Lmk!

tboerger:
@bradrydzewski is this still required?

jrm2k6:
@bradrydzewski this seems to be still needed based on my looking around in the current codebase. It looks like it still shows the author instead of the person triggering the promotion in the UI, which is confusing.

Tejukapare:
Why is this not merged yet?

ste93cry:
Is there any plan to finish the work and merge this? I'm using Drone at work with promoted builds and we're finding confusing a lot to see the author of commit the build is running on instead of the person that triggered the promotion...

ste93cry:
We're using Drone at work with promoted builds and it's quite confusing to see the author of the commit the build is running on instead of the person that promoted the build... As far as I see all comments that had a meaning are resolved, so @lnikell can you please review this?

bradrydzewski:
I believe this was resolved by https://github.com/drone/drone-ui/commit/a3ce793ddcca31385c0c5c525ac2c8e752f1f793.  Please try upgrading to a newer version of Drone. I am going to close this because it cannot be merged; this pull request was authored prior to a complete rewrite of the user interface.

PR #287, opened 2019-05-03
Add button to show all log lines

dramich:
Currently with large logs it takes many clicks to be able to see the
entire log file or the top of the log file.

This button displays all log lines with one click.

### DIFF
--- a/src/views/Build.vue
+++ b/src/views/Build.vue
@@ -129,6 +129,12 @@
             <Loading v-if="logsShowState === 'loading'"/>
 
             <div class="output-content-actions" v-if="moreCount">
+              <Button size="l" outline borderless class="output-button" @click.native="handleAll">
+                Show all lines
+              </Button>
+            </div>
+
+            <div class="output-content-actions" v-if="moreCount">
               <Button size="l" outline borderless class="output-button" @click.native="handleMore">
                 Show {{Math.min(moreCount, logStep)}} lines more
               </Button>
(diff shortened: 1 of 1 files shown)


bradrydzewski:
thanks, can you also provide a screenshot so that design team can review and provide feedback?

dramich:
Yes, see below:
Screen Shot 2019-05-03 at 11 28 06 AM

bradrydzewski:
thanks! cc @lnikell

lnikell:
@bradrydzewski two blocks with show logs will be too much, I would prefer to put it near with download button or think about some option to keep show all and 250 in one line. We will come up with some solution.

stevecrozz:
Chiming in with another idea, it might be cool to do something similar to what slack does and anchor you to the bottom when you first visit a room (or a stage in our case) but automatically load more as you scroll back.

bradrydzewski:
The reason we do not display all logs by default is because logs greater than 2mb in size tend to lock up the browser. This was a significant issue in older version of drone and is the reason for the current user interaction, which clicking a button to reveal more logs is required.

The reason for using a button is simplicity. We could implement more complex user interactions, but the complexity has a cost in terms of development effort and maintenance. The benefits of the complex interaction has to justify the cost. In this case, I think a simple approach is both reasonable and sufficient.

Note that I do appreciate the suggestion, even though I prefer a different approach :)

dramich:
@lnikell Any thoughts on this? I am happy to make the changes.

PR #302, opened 2019-07-22
Fixes shoulder surfing vulnerability in secrets module

rewanth1997:
### Issue -
The user can store secrets in drone UI but the secrets are **"displayed in plain text"** while the user types the values into the input field.

### Fix -
This PR sets the input to "password type" to make sure the sensitive information isn't visible to others through shoulder surfing.

cc: @hrushikeshk


bradrydzewski:
the password field does not allow for multi-line inputs, which is important since a large percentage of secrets are multi-line strings (private keys, json documents, etc)

rewanth1997:
Valid point! But I don't think this CSS feature will work in all browsers. So, this wouldn't be the best solution.

I think we can use JS/Jquery to hide every character entered into the input field into dot/star character. Do you think this is a good way to solve this issue?

tboerger:
Maybe build a widget where the content is hidden by default via js including an overflowing icon to toggle the visibility?

bradrydzewski:
In my experience hidden textboxes that intercept keystrokes, mouse clicks, text selection, etc are highly error prone and would become a maintenance burden. I do not want to introduce such complexity into the codebase. I can already see the issues piling up :)

I would be willing to accept a pull request for `-webkit-text-security: disc` which data suggests would support 94.4% of users or more. The remaining 4.6% of users can fallback to the command line if they feel uncomfortable with the unmasked textarea.

bradrydzewski:
As a longer term solution, perhaps we can look into https://github.com/noppa/text-security which is a special font that renders all characters as discs. Since it is a font it would be cross-browser compatible.

rewanth1997:
Hi @bradrydzewski @tboerger ,

The text-security solution ran successfully in Edge, Firefox and Chrome. So far we think the text-security module is the best solution to prevent shoulder surfing attacks.  Please review the PR and share your views.

tboerger:
IMHO it should rely on https://www.npmjs.com/package/text-security instead of bundling this vendor files.

rewanth1997:
I presumed that the font will be stable at all times and cloned "text-security" GitHub repo into the code instead of using npm package. After you mentioned about npm bundle, even I felt using npm is the best way to integrate new modules into the code. Please have a look at the latest commit.

tboerger:
I just realized now that you are targeting the wrong branch, the react app isn't used anymore, the drone ui is using the vue branch.

rewanth1997:
Thanks for pointing that @tboerger

PR #304, opened 2019-07-28
Integrates text-security npm module to Secrets component

rewanth1997:
Hides the sensitive secret information while typing as password disc symbols to prevent shoulder surfing.

@bradrydzewski @tboerger There were some compilation errors when I tried to include svg, eot and ttf font files. It seems they require updated version of file-loader and url-loader modules. ( Ref: https://github.com/webpack/webpack/issues/6419 ). The latest version of both the modules are added to package.json

Closes #302


src/components/editable-list/Secrets.vue
```diff
@@ -86,6 +86,11 @@ export default {
 </script>
 
 <style scoped>
+@import "../../../node_modules/text-security/dist/text-security-disc.css";
```
tboerger: You should be able to import ~text-security...

rewanth1997: No, I guess tilde (~) paths doesn't work with import statements. At least it didn't work here.

tboerger: I'm not sure how the syntax is exactly, but since this project is using vue-cli and because of that webpack it should be supported. I'm pretty sure @lnikell can say something about that.

IMHO it's bad to import it like that.

rewanth1997:
@tboerger I wasn't familiar with VUE, so I faced issues in importing node_modules. After digging up a bit, I found a new way and its working like a charm. Please check it.

rewanth1997:
@bradrydzewski @tboerger any changes required for the code? any updates?

rewanth1997:
> I would name the class `hide-secret`, but besides that LGTM

Done. I was more associated with JS, so just got used to JS coding conventions.

tboerger:
cool :+1:

till:
The change looks great, is anyone merging this soon?

PR #305, opened 2019-08-12
Truncate commit message to first line in RepoItem

dacruz21:
RepoItems attempt to display the entire commit message for each build
triggered by a commit. This is okay for most users as the message will
run out of space and be truncated. However, I among many others with
Hi-DPI displays have customized Drone's CSS to take up more monitor
space, which leads to some very ugly build names, as the RepoItem
displays the commit message and body.

This PR trims the title to just the first line of the commit message.

### DIFF
--- a/src/components/RepoItem.vue
+++ b/src/components/RepoItem.vue
@@ -14,7 +14,7 @@
       <div class="header" :title="title">
         <span class="title">
           <span class="number" v-if="number">#{{ number }}. </span>
-          <span>{{ title }}</span>
+          <span>{{ title.split('\n')[0] }}</span>
         </span>
 
         <Button outline borderless


bradrydzewski:
thanks! any chance you could provide a before / after screenshot so we can see the benefit?

dacruz21:
Before:
image
After:
image

bradrydzewski:
I see what you mean, thanks for providing the screenshots!

I wonder if you think we should additionally limit the length to 72 characters? I believe this is recommended max for the first line of the commit.

dacruz21:
The standard for commit message length is usually 50 characters for the title, and the body word-wrapped at 72 characters. I can trim it to max 50 characters as well if you'd like, but (personally) I've always found it a little annoying when GitHub truncates my commit message if I provide too long of a title. Thoughts?

bradrydzewski:
We can merge this as-is and decide on overall column length at a later time. I agree that it can be annoying to truncate the message, especially at 50 characters. This change will be included in the 1.3 release.

PR #317, opened 2019-11-12
Add Ability to Filter Builds by Branch

mlclmj:
This PR is an initial pass at adding some filtering abilities to the builds list.

This keeps the same default functionality of the current build list by showing all recent builds, however adds the ability to filter that down further by branch.
A dropdown lists all the branches available to filter by, sorted by the frequency of builds (more builds → higher on the list), floating the `master` branch to the top of the list if it's present.

I'm very open to making changes to how this works, but it's something we're starting to use internally and seeing as though it's something our users have been asking for, I assume it's a feature others would like to see in Drone as well.

Thanks!

### DIFF
--- a/src/views/Builds.vue
+++ b/src/views/Builds.vue
@@ -132,9 +132,6 @@ export default {
     },
     shrinkBuild(build) {
       return { ...build, message: null };
-    },
-    input(branch) {
-      this.defaultBranch = branch;
     }
   }
 };
--- a/src/views/Builds.vue
+++ b/src/views/Builds.vue
@@ -87,9 +87,8 @@ export default {
       return builds;
     },
     branches() {
-      const counts = {};
-      let branches = new Set();
-      branches.add("(All)");
+      const counts = {"(All)": 0};
+      let branches = new Set(["(All)"]);
       this.builds.forEach( build => {
         if (typeof counts[build.target] === 'undefined') {
           counts[build.target] = 0;
@@ -100,7 +99,13 @@ export default {
       branches = [... branches];
       // Sort branches by most builds but prefer master if it's in there
       branches = branches.sort((a, b) => {
-        if (b === "master") {
+        if (b === "(All)" && a === "master") {
+          return 1;
+        }
+        if (b === "master" && a === "(All)") {
+          return -1;
+        }
+        if (b === "master" || b === '(All)') {
           return 1;
         }
         return counts[b] - counts[a];
--- a/src/views/Builds.vue
+++ b/src/views/Builds.vue
@@ -8,6 +8,12 @@
         <div class="empty-message">Your Build List is Empty.</div>
       </Alert>
 
+      <div class="build-filter" v-if="!showEmptyListAlert">
+        <Card>
+          <label class="control-label">Branch</label> <BaseSelect :options="branches" v-model="defaultBranch"/>
+        </Card>
+      </div>
+
       <router-link
         class="build"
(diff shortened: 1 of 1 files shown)


bradrydzewski:
@mlclmj sorry for the delay on this.

We recently added a new branches page which shows a summary (below) but I still think there is value in being able to filter builds.

Screen Shot 2020-03-19 at 9 13 16 AM

Instead of a dropdown, I was hoping we could add a filter button:

Screen Shot 2020-03-19 at 10 41 12 AM

Upon clicking the button we would be presented with filter options. I am not sure if this would be a menu or a popup. I like whatever option is the most simple :)

Screen Shot 2020-03-19 at 9 38 53 AM

One consideration is that we may want to, in the future, expand the interaction to filter by event type, which is why I'm thinking this would be a popup or menu as opposed to a `<select>`

tonglil:
What about two separate selects/drop downs? My preference is dropdowns because they are more accessiblility friendly (think using keyboard and tabbing through options). 

I can see a combination of filtering by event x branch: all PRs to master, all pushes on master, all PRs, all pushes, all events on master, etc...

Personally I would find more value having the feature in a release, and then iterating on it - vs waiting until it covers all/future use cases.

bradrydzewski:
> What about two separate selects/drop downs? My preference is dropdowns because they are more accessiblility friendly (think using keyboard and tabbing through options).

I could see us doing something like dribble, where the filter button is a toggle that reveals dropdowns to filter the list.

Screen Shot 2020-03-19 at 12 25 25 PM

> I can see a combination of filtering by event x branch: all PRs to master, all pushes on master, all PRs, all pushes, all events on master, etc...

Filtering will be limited to match our API endpoints and database indexes. While we can filter pretty much anything in-memory, we really need to augment filtering with API calls. Consider the following scenarios where in-memory filtering will fall short:

Scenario 1:

_User wants to filter the list to display all builds for the "develop" branch. The repository is very activate and there are no build entries for branch "develop" in the in-memory list. Therefore "develop" is not displayed in the dropdown._

Scenario 2:

_User wants to filter the list to display all builds for the "develop" branch. There are dozens of builds for branch "develop" but only 1 entry exists in the in-memory list. The user filters based on branch "develop" and it therefore only displays 1 entry._

mavimo:
What do you think to use an approach "a la sentry.io"? I think they have nice usability, easy extensibility and will be really easy to add new filter or rules without change UI too much.

bradrydzewski:
I prefer to keep things simple. We tested a complex filter widget similar to sentry in 0.8.  It was buggy and created maintenance burden.  Simple dropdowns for event and branch (and maybe author) should be good enough for our users and would give us feature parity with other popular systems (Travis, Circle, etc).

tonglil:
I am cool with what you suggested with regards to the filtering limitations.

I would like to say that while hiding filters by default is behind another button is "clean", for the lack of a better word it is "annoying" to use.

I believe the filters will be used frequently, so having to click on a button to show the dropdowns is just an extra click each time.

 Imagine you had to click a button before being able to select from this dropdown on github: 
image

bradrydzewski:
Thanks for the feedback. We intend to solve this by remembering whether or not the filter section (similar to dribbble screenshot above) is toggled or not (e.g. via localstorage).  Some users will want it always visible and others won't.  I think this strikes a good balance to satisfy both camps.

tonglil:
That's a good design. Thanks!

Issue #314, opened 2019-10-19
Dark Mode/Theme


A-UNDERSCORE-D:
The default interface is pretty eye-searing for anyone that is either working at night in a dark room with dark mode on all their IDEs and other dev tools, or simply prefers light-on-dark text. A quick toggle to switch the UI to a dark mode would be wonderful.

PR #319, opened 2020-01-02
Support an ability to switch the UI to a dark mode

pirsquareff:
According to #314, this PR adds an ability to switch the UI to a dark mode.

Here are screenshots of the main UI.
drone-dark-mode-demo
home-dark
build
repo-settings
user-settings

## Changes
- [x] Colors for the dark theme (`src/assets/styles/_variables.scss`)
- [x] SCSS mixins for theming (`src/assets/styles/_mixins.scss`)
- [x] Adjust all components' style to support the dark mode
- [x] A control component to switch the UI in the UserMenu
- [x] Lib for updating/fetching theme to/from Local Storage (`src/lib/theme.js`)

This keeps the default UI the same. All colors are properly adjusted for the dark mode. In order to support a dynamic theme switching, two SCSS mixins have been introduced in this PR: `themed()`, `themed-only()`.

For example, if we want to add support for the dark theme to a following SCSS selector
```scss
.selector {
  color: $color-text;
}
```
We can wrap it with the `themed` mixin and get the color by a `tget()` function.
```scss
.selector {
  @include themed {
    color: tget("color-text");
  }
}
```

With colors defined separately in `_variable.scss`, the above code will be transpiled to CSS as follows:
```css
.theme--default {
  .selector {
    color: <color-text-for-default-theme>;
  }
}

.theme--dark {
  .selector {
    color: <color-text-for-dark-theme>;
  }
}
```

I hope this might be a cool feature for Drone UI. If there is any suggestion or improvement that I can contribute to, please let me know.

Thank you.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,6 +2,16 @@
   <router-view />
 </template>
 
+<script>
+import { applySavedTheme } from "@/lib/theme";
+
+export default {
+  mounted() {
+    applySavedTheme();
+  }
+};
+</script>
+
 <style lang="scss">
 @import "./assets/styles/reset.scss";
 @import "./assets/styles/global.scss";
--- a/src/assets/styles/_mixins.scss
+++ b/src/assets/styles/_mixins.scss
@@ -30,3 +30,44 @@
     @content
   }
 }
+
+@mixin themed($attached: false) {
+  @each $theme, $map in $themes {
+    $wrapped-selector: ".theme--#{$theme} &";
+    @if $attached {
+      $wrapped-selector: "&.theme--#{$theme}";
+    }
+    #{$wrapped-selector} {
+      $theme-map: () !global;
+
+      @each $key, $value in $map {
+        $theme-map: map-merge($theme-map, ($key: $value)) !global;
+      }
+
+      @content;
+      $theme-map: null !global;
+    }
+  }
+}
+
+@mixin themed-only($theme: default, $attached: false) {
+  $wrapped-selector: ".theme--#{$theme} &";
+  @if $attached {
+    $wrapped-selector: "&.theme--#{$theme}";
+  }
+  #{$wrapped-selector} {
+    $theme-map: () !global;
+    
+    $map: map-get($themes, $theme);
+    @each $key, $value in $map {
+      $theme-map: map-merge($theme-map, ($key: $value)) !global;
+    }
+
+    @content;
+    $theme-map: null !global;
+  }
+}
+
+@function tget($key) {
+  @return map-get($theme-map, $key);
+}
--- a/src/assets/styles/_variables.scss
+++ b/src/assets/styles/_variables.scss
@@ -5,29 +5,91 @@ $bp-mobile: 480px;
 /* [-----------](-----------)[-----------> */
 /*           480px       980px             */
 
+$surface-color: #ffffff;
+$surface-color--dark: #161616;
+
 $body-color: #f7f8fa;
+$body-color--dark: #0a0a0a;
+
 $body-background-size: 100% 100%;
 
 $color-text: #1e375a;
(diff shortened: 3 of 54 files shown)


tboerger:
The Screenshots are really looking awesome

bradrydzewski:
thanks for taking the time to submit a pull request.  We are planning to use server side rendering for some pages (mostly admin pages and other pages that do not require live updates) so I'm not sure we want to use a toggle or store this value in local storage. Instead perhaps we should detect whether or not the host OS uses dark mode.

pirsquareff:
@bradrydzewski Great to hear your thought on the plan. Perhaps it can be implemented in a way that a theme defaults to the host OS mode, but still, let a user change and stick to that preference thereafter. Moreover, I think the theme should not be global across all the user's browser. That is why I decided to use local storage in this case. 

Anyways, I'm very open to suggestions. Keep me posted about the next step.

bradrydzewski:
> I think the theme should not be global across all the user's browser.

agreed

> Perhaps it can be implemented in a way that a theme defaults to the host OS mode, but still, let a user change and stick to that preference thereafter

agreed this would be ideal, but I think the implementation is tricky when you consider server side renders.  Perhaps a cookie would be a better approach since this could be read by the server and used to return the appropriate markup. Just thinking out loud ...

pirsquareff:
@bradrydzewski Ah I see your point regarding the server-side renders, a cookie would be a better approach in this case as you said. 

Still have a question about changing a theme on the fly, should both default and dark theme (CSS) be bundled together as in this implementation so that when a user changes it, a page doesn't have to be reloaded? Or is there any way to serve them separately from the server-side (not sure about the implementation here)?

tboerger:
@bradrydzewski server-side rendering shouldn't be a problem since this only changes the css beside the base dark theme class assignment.

briancurt:
Please don't let this die, it looked great 🙏

tuxity:
Looks nice! when it will be merged?

justinhauer:
@bradrydzewski plz don't let this one die, would love a dark mode option

infinytum:
Any plans to merge this? I'd love to have this!

Cryoge:
This would be super nice!

kushwahashiv:
is that still alive and going well? whats the timeline if answer is yes?

unrealcloud (approved):
🕶️

mrjuan1:
Any plans to eventually merge this?

infinytum:
@bradrydzewski Could this be merged please? Our eyes demand darkness! :)

adyanth:
Any chance on this merging soon? Looks really good! @bradrydzewski

glaszig:
2.5 years. lgtm. all the white is really annoying.

craftedsystems:
I was annoyed too and created this bookmarklet.
`javascript:(function()%7B var style %3D document.createElement(%27style%27), styleContent %3D document.createTextNode(%27html %7B filter: invert(.85) %7D %5Bclass%5E%3D%5C"sidebar_wrapper%5C"%5D, %5Bclass%5E%3D%5C"log-view%5C"%5D %7B filter: invert(1) %7D %27)%3B style.appendChild(styleContent )%3B var caput %3D document.getElementsByTagName(%27head%27)%3B caput%5B0%5D.appendChild(style)%3B %7D)()%3B`

sdooweloc:
Crazy that a product geared towards devs and such that often prefer to keep their eyes in their sockets doesnt have a dark mode. Any chance we could get one at some point? The searing white kills me any time I go to my tab for drone.

glaszig:
i consider drone abandonware now. i smelled that since the harness takeover. in the meantime, until i can manage to migrate my entire ci to another oss solution, i'm using the booklet above.

PR #322, opened 2020-01-31
Force close of search box after item selection

### LINKED ISSUE #263: Grey overlay after selecting repo to activate
**Version**: 1.0.0-rc.5
**Description**:
After selecting a repository to activate the overlay stays gray and the user must refresh the page to be able to click **Activate**
activate

ViViDboarder:
In Firefox, for some reason selecting an item with the mouse would not
clear the search overlay. It appears that it would generally clear when
the blur function is called, but something appears to be different when
selected by mouse vs enter key.

This forces a call to `this.close()` to close the dialog without relying
only on the blur action.

Fixes #263

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -95,6 +95,7 @@ export default {
     onItemSelect() {
       this.$refs.searchInput.$el.blur();
       this.clear();
+      this.close();
     },
     clear() {
       this.query = "";


ViViDboarder:
Is anyone able to review this? It's a pretty small diff. :)

PR #326, opened 2020-03-12
Add Pulls and Pushes view to individual repo

NicsWorld:
This PR adds a filtering of the Activity feed by pull requests and pushes to each individual repo view.

Some questions I wanted to clarify - 

1. There were existing lint errors, I cleaned up a few that I saw but there are still some lint errors, should I clean all of them up?
2. The readme calls out running tests, but I could not find any tests. Am I missing them?

Example:

drone-ui

### DIFF
--- a/src/components/RepoItemLabel.vue
+++ b/src/components/RepoItemLabel.vue
@@ -45,20 +45,16 @@ export default {
       }
     },
     hrefPR() {
-      return this.build && this.repo &&
-        `/link/${this.repo.slug}/tree/${this.build.ref}`;
+      return this.build && this.repo && `/link/${this.repo.slug}/tree/${this.build.ref}`;
     },
     hrefTag() {
-      return this.build && this.repo &&
-        `/link/${this.repo.slug}/tree/${this.build.ref}`;
+      return this.build && this.repo && `/link/${this.repo.slug}/tree/${this.build.ref}`;
     },
     hrefBranch() {
-      return this.build && this.repo &&
-        `/link/${this.repo.slug}/tree/refs/heads/${this.build.target}`;
+      return this.build && this.repo && `/link/${this.repo.slug}/tree/refs/heads/${this.build.target}`;
     },
     hrefCommit() {
-      return this.build && this.repo &&
-        `/link/${this.repo.slug}/commit/${this.build.after}`;
+      return this.build && this.repo && `/link/${this.repo.slug}/commit/${this.build.after}`;
     },
     branch() {
       return this.build.target;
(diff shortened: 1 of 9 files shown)


bradrydzewski:
Hey @NicsWorld we should touch base in Gitter to discuss.

We have some code in the `vue-1.0.0` branch for dedicated branch and deployment screens. You can see a screenshot here:

Screen Shot 2020-03-13 at 6 20 43 PM

The above interaction is more comparable to the GitHub branches and pull request views and gives teams a birds eye view of which builds, deployments and pull requests are passing vs failing.

Screen Shot 2020-03-13 at 6 31 08 PM

We could then provide an option to drill-down and view all branches. Travis has an interesting grid view although I would concerned with database performance when rending this page ...

Screen Shot 2020-03-13 at 6 33 13 PM

Lastly, we now have dedicated endpoints for fetching builds by branch. This is preferable because you are otherwise sourcing this data from the list of builds in the Vue store, which is going to be limited to the last 25 builds.

jones2026:
@bradrydzewski would now be a good time to maybe cleanup some of the old branches so this is less confusing in the future?

ashwilliams1:
@bradrydzewski I think we can probably migrate to master now right?  @jones2026 I believe the reason for separate branches is that 0.8 used `go get` to pull the UI from master and for a period of time we were maintaining 0.8 even after 1.x was launched.  Now that 0.8 is formally end of life I think it is safe to move the 1.x code to master.  Do you agree Brad?  If so I can make the change.

bradrydzewski:
Yes agreed, I went ahead and moved the `vue` branch to `master`. I will spend some time today trying to split the changes in `vue-1.0.0` into separate feature branches so they can be worked on independently going forward.

bradrydzewski:
We moved some of the code from `vue-1.1.0` into master including the branch view (pictured below). You can see the interaction live [here](https://cloud.drone.io/drone/hello-world/branches).

Screen Shot 2020-03-19 at 9 13 16 AM

Instead of having separate tabs to show all builds by branch, pull request, etc I was thinking we could provide an option to filter the main activity screen. Ideally the interaction would be really simple, for example:

Screen Shot 2020-03-19 at 9 38 53 AM

NicsWorld:
Hey @bradrydzewski thanks for all the info! II will shoot you a message on gitter shortly. I actually spiked out a filter menu similar to what you showed, i'd love to pursue or help anyway I can!

bradrydzewski:
Also see https://github.com/drone/drone-ui/pull/317 which filters by branch. This could be expanded to filter by event type.

PR #329, opened 2020-04-10
Deployments and deployment action

imranismail:
This implements the deployments tab and deployment action using the new APIs provided by 1.7.0

image
image
image

Some additional features that I'm thinking of including in this PR would be autocomplete on the target environment input field based on past deployment builds.


src/views/Build.vue
```diff
@@ -9,6 +9,43 @@
           <IconSource/>
         </Button>
 
+        <div style="position: relative; display: inline-block">
```
imranismail, line 12: Missed this bit, I'll fix it in a bit

Rush:
Nice, would it be possible for `to` to be a select box based on defined environment in YAML?

imranismail:
> Nice, would it be possible for `to` to be a select box based on defined environment in YAML?

There is no API currently available to do so.

But there is a way to view past deployments and I might be able to make an autocomplete input field based on past deployments data.

The downside is: you'd still have to manually trigger the deployment once before it's available in the autocompletion list.

Rush:
There are some autoselect libraries which open a popup on initial focus. Then it could double down as an input field AND a select.

imranismail:
> There are some autoselect libraries which open a popup on initial focus. Then it could double down as an input field AND a select.

I'm still looking for a Vue component that could do both.

This is my first time writing Vue code and I'm not well versed with which Vue component is the defacto standard for autocompletion.

Could you share some if you found any?

Rush:
https://vuetifyjs.com/en/components/autocompletes/
Seems this opens on focus
image

imranismail:
@Rush I did a quick hack to put together the autocompletion and build parameters

image
image

src/components/forms/DeploymentForm.vue
```diff
@@ -0,0 +1,213 @@
+<template>
+  <BaseForm @submit.native.prevent="handleSubmit" class="form">
+    <header>
+      <h2 class="title">
+        Deploy #{{buildNumber}}
```
bradrydzewski, line 5: we should remove the term "deploy" since this tends to confuse people.  This is why we change the term "deploy" to "promote" in 1.0 so that it was less confusing

src/views/Repo.vue
```diff
@@ -38,6 +38,7 @@
     <nav v-else-if="showTabs">
       <router-link :to="'/'+slug" :disabled="!repo.active">Activity Feed</router-link>
       <router-link :to="'/'+slug + '/branches'" :disabled="!repo.active">Branches</router-link>
+      <router-link :to="'/'+slug + '/deployments'" :disabled="!repo.active">Deployments</router-link>
```
bradrydzewski, line 41: rename to "Promotions"

bradrydzewski:
Thanks for sending the pull request!  Can we split these into separate, smaller pull requests so that we can more quickly get them reviewed and merged?  Here is how I recommend splitting them:

1. pull request for "promotions" tab
2. pull request for "promotion" popup with simple text box
3. pull request for "promotion" popup with parameters / autocomplete / autofill

src/components/forms/DeploymentForm.vue
imranismail: This form is being used for rollback as well, do I dynamically change the title based on the `promote/rollback` event?

bradrydzewski: yes, that sounds like a good idea, for example

> Promote #42 

and

> Rollback to #42

bradrydzewski: actually, you may be able to remove the build number completely from the popup.  Presumably a user knows which repository and build they are promoting, and the build number should also be in url bar and title.

imranismail: @bradrydzewski what about the button label, should I name it "Environments"

kirillta:
@imranismail are you planing to rearrange that PR?

smoulderme:
Any updates?

brandoncollins7:
I tried building this ui and pointing it to my server on kubernetes but for some reason I keep getting 401 on the call to api/users, even though when I visit that url in the browser I get a valid response back. Any ideas what could be happening there?

bradrydzewski:
@imranismail I was not able to merge the PR as-is but was able to cherry pick many of the changes, and tweak where needed.  However, you really deserve all the credit here and I wanted to thank you for your contribution. I will make sure you get all the credit in our release announcement :)

kirillta:
@bradrydzewski why didn't you use opened PR #332 Will you pay attention to it, or I could close it as well?

ashwilliams1:
this pull request was revived (with edits) and is available in master

PR #335, opened 2020-07-03
Show build params on build page

dschmidt:
This PR adds a simple table showing build parameters:

Screenshot_20200703_174117

### DIFF
--- a/src/components/RepoItem.vue
+++ b/src/components/RepoItem.vue
@@ -51,6 +51,14 @@
         </div>
       </div>
     </div>
+    <div v-if="showParams && !!build.params" style="margin-top: 20px;">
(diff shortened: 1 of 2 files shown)


bradrydzewski:
I would like to handle this a little differently.  I would like a popup where we can display a table of all build fields.  This is a more scalable solution and prevents us from having to get a designer involved every time we want to add more fields to the user interface.

dschmidt:
What kind of popup do you think of? One that opens when you click something and stays open or one that just gets shown when you hove something? The latter makes it hard to copy values from the table then.

tboerger:
I think he is talking about a modal on click.

PR #338, opened 2020-08-13
Merge build logs instead of concatinate it

wingyplus:
Sometimes server sent build log event since first line and drone-ui
append to the old build log. This approach make a user see a build since
first line again even it already shown. Fix it by merge a new logs to
the old logs by using field "pos" as a array index.


wingyplus:
@bradrydzewski Love to see if you've any suggestions. :)

Issue #309, opened 2019-09-10
Text goes outside the log box


GCorbel:
With some steps, the text goes outside the box in logs as you can see here : 

![image](https://user-images.githubusercontent.com/1296201/64617375-5a0abe80-d3ac-11e9-9fd7-a5c93c0fca58.png)

It's written white on white so we have to change for the full screen view or download logs.

bradrydzewski:
fixed by dc6cbfb66b3b4449eabd76a7431150a8d7de4895. Updated distribution will be included in the 1.5 release of Drone.

PR #341, opened 2020-10-15
use pre-line replace pre-wrap

HuberTRoy:
`pre-wrap` will keep all spaces that will cause text to exceed the log box like #309 .

`pre-line` includes `pre-wrap` but the sequences of whitespace will collapse into a single whitespace.

### DIFF
--- a/src/views/Build.vue
+++ b/src/views/Build.vue
@@ -789,7 +789,7 @@ $output-header-sticky-offset: 20px;
 
 .ol-html {
   word-break: break-word;
-  white-space: pre-wrap;
+  white-space: pre-line;
   padding-right: 15px;
 }
 


silverwind:
I think `overflow-wrap: anywhere` is the proper fix for such overflows.

HuberTRoy:
@silverwind 
`word-break: break-all` will split one word to two separate part, it seems not good for this case.
`overflow-wrap: anywhere`  maybe newest than now. https://caniuse.com/?search=overflow-wrap

Thanks for `overflow-wrap` knowledge, I dont know this before.

silverwind:
> will split one word to two separate part, it seems not good for this case

That's how terminals do it generally. At least the ones I use.

silverwind:
Note that `pre-line` will remove leading whitespace on lines which is not desirable.

`pre-wrap`:

image

`pre-line`:

image

I guess `overflow-wrap` might be too new of a property overall, so `word-break: break-all` is probably the right solution, thought I didn't really observe breaking issues like you show, even with long lines without any whitespace on them. Can you post a string that will reproduce the issue?

HuberTRoy:
@sliverwind

It's a rare case. It seems the reason of this case is the whitespace too long.

```
Total num: 560, size: 11,485,153. Dealed num: 311(upload 308 files, 3 directories), OK size: 7,614,159, Progress: 66.296%, Speed: 1485.62KB/s
                                                                                                                                              
Succeed: Total num: 560, size: 11,485,153. OK num: 560(upload 509 files, 51 directories).
```

`word-break: break-all` and `overflow-wrap: anywhere` didn't slove this problem I just now test...

silverwind:
Looks fine to me if I paste into the `ol-html` nodes:

image

HuberTRoy:
@silverwind 

Weird... It should be one line not three.

silverwind:
Seems the issue is Chrome-only, it does work fine in Firefox.

Either it's a Blink bug or some strange handling regarding tables and overflow. I can't get the line to break using any of the regular means that normally work outside of `<table>`.

silverwind:
Adding `display: flex` to `ol-html` seem to fix it for me.

HuberTRoy:
@silverwind 
Yes, it's Chrome-only and I agree with you and the case of use `pre-wrap` is good than `pre-line`.

ahh. It seems no solution for both case now in `<tr>`.

It's not good for me to add `display: flex`.

`Chrome  86.0.4240.198 64bit`.

silverwind:
Strange, certainly works for me on Chrome 86, macOS.

image

HuberTRoy:
Sadly= =...and luckly.. The case is not very frequently and not unacceptable.

I think the PR should be close for now.

Thanks your response.

Issue #298, opened 2019-07-03
Add a new build button to the repository screen


Skaronator:
Hi,

it would be great to have a "New Build" button in the repository screen. Clicking on it should open a modal which ask for a 

- *name* of that build
- *Commit* which should be built
- *Branch* of that commit
- *Environment Variables* a text area where it should be possible to define additional env variables

Buildkite has this option and would like to see that in drone as well.

![image](https://user-images.githubusercontent.com/2886913/60582615-7295a180-9d89-11e9-81fe-c0f80aa7cc93.png)

![image](https://user-images.githubusercontent.com/2886913/60582629-7b867300-9d89-11e9-8150-78ea64e9158e.png)

avierr:
Yeah I hate it when my build artifact upload fails due to network issues and I have no way to re-upload/re-build them.
I have to again go through a fake commit and then PR process just to make this build work again.

Stuck with Jenkins because of the above feature not being present.

tboerger:
@avierr on network issues you can always restart a build which will launch a new build with the exact same parameters.

stixx:
When is this feature coming on the roadmap?

toopay:
It should be straightforward. I can start the effort if its on the owner of this repo's interest

fabiosussetto:
I think this would be a very welcome addition, useful for a lot of people w/o too much dev effort.
I can have a look at implementing this in a PR as well, if @bradrydzewski would be up for this functionality.

Skaronator:
This would be a really powerful feature when using over the Drone REST API. You could create your own Dashboard to trigger new builds which do whatever you want. For example a deployment on a special environment.

Akkadius:
@toopay I look forward to it!

shamimgeek:
This would be really great feature. expecting sooner.

A-limon:
any update?

toopay:
@Akkadius I haven't heard anything from the maintainer, so I reluctant to start anything.

We're actually using Spinakker atm :grinning:

mritd:
Is there any update?

sbussard:
This would be a really useful feature, really an expected feature to be honest. Drone is great though btw

0xMH:
And if possible adding a field for `event type` that will make it possible for us to change/specify the behaviour of the pipelines by handling the event they receive.

ivmilicevic:
[...]

PR #343, opened 2020-10-20
Add a 'New Build' button and modal on builds page.

youcandanch:
Many other CI systems -- Jenkins, Buildkite, Concourse, etc. -- include a way to kick off an arbitrary build from the UI by just providing a branch (and optionally a commit hash). This is helpful for multiple reasons, and is a fairly low lift add.

drone

One thing I didn't do -- primarily just lacking time right now -- is add in a way to populate environment variables specific to the run. Would be easy enough to add another form field and then add them as querystring params.

Resolves https://github.com/drone/drone-ui/issues/298.

### DIFF
--- a/src/actions/build.js
+++ b/src/actions/build.js
@@ -85,15 +85,44 @@ export const cancelBuild = async ({ commit }, { namespace, name, build }) => {
   }
 };
 
+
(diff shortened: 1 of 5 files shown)


tphoney:
Hey @youcandanch thanks for the PR, we drew inspiration from this PR for V2 of the UI. Thank you for your patience.
image

PR #347, opened 2020-12-13
Enable collapsing/expanding of stages

LukasHirt:
This PR introduces the ability to collapse/expand stages. IMHO it makes sense to enable users to do this in case the user has a pipeline with a lot of steps which makes it then "more work" to scroll between different pipelines as he needs to first scroll through all those steps. When selecting the pipeline, it is of course always expanded. Same when switching between pipelines.

For accessibility, I've implemented it as a `button`. I did it directly in the `Stage` component and I think it would make sense to move there also the `router-link`. This would save some code in the `Build` component and would make `Stage` really encapsulated. This is why I created the styles for the button in the `Stage` component and didn't reuse the ones in `Build`. Let me pls know what you think about this idea @bradrydzewski 🙂 If you'd like this approach I could move it in there already in this PR or in a separate one - your call 😉 

When using the `button`, it created one small change in behaviour though. Since `button` is getting outline by default when clicking on it, I disabled it via CSS. This breaks the behaviour of the other pipelines though - since they are links, you do not get the outline when clicking but you get it when tabbing. This is now not working for the selected pipeline.

As one last tiny change, I added small padding of 5px to the title of the pipeline on hover/focus. This is to let the chevron icon on the right breath a little since previously the text would go all the way to that icon (the three dots when the text is truncated).

### DIFF
--- a/src/components/Stage.vue
+++ b/src/components/Stage.vue
@@ -1,15 +1,17 @@
 <template>
-  <section :class="{ stage: true, 'has-steps': hasSteps, 'has-content': hasContent}">
-    <header>
-      <Status :status="stage.status"/>
-      <span :title="stage.name">{{ stage.name }}</span>
-      <time-elapsed v-if="stage.started" :started="stage.started" :stopped="stage.stopped">
-        <Hint position="top" align="right" showOn="hover">Full stage duration</Hint>
-      </time-elapsed>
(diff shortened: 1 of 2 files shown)


LukasHirt:
One last note - I hope I chose the correct branch. I didn't find any contribution guidelines where it would say which one to choose so I chose one that I saw in other PRs :wink:

LukasHirt:
Closing as not relevant anymore with the new UI

PR #356, opened 2021-05-28
feat(Builds): Add filter by event

julienduchesne:
Allows filtering repo builds by push, cron or PR

The pagination doesn't adapt to the new type (you get a list of less than 50 items), you may have to click a few times on show more to get a full 50 items list (depending on your build event type distribution). However, I really think that this is a really useful feature. There is currently no way to differentiate a cron from a normal build (or to get the average duration of all crons)

Here's what it looks like:

Default filter (all):
image

With cron filter:
image

### DIFF
--- a/src/pages/builds/builds.jsx
+++ b/src/pages/builds/builds.jsx
@@ -1,11 +1,12 @@
 import classNames from 'classnames/bind';
 import React, {
-  useLayoutEffect,
+  useLayoutEffect, useMemo, useState
 } from 'react';
 import { useHistory, useRouteMatch } from 'react-router-dom';
 
 import BuildList from 'components/pages/repo/build-list';
 import Button from 'components/shared/button';
+import Select from 'components/shared/form/select';
 import ZeroState from 'components/shared/zero-state';
 import { useCustomTitle } from 'hooks';
 import { useBuilds } from 'hooks/swr';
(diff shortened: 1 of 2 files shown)


julienduchesne:
@bradrydzewski

tphoney:
Hey @julienduchesne thanks for the suggestion and the PR. Unfortunately we would be unable to accept this request like that at this time.  There are a few reasons for this. Firstly as you suggest the caveat is a problem, we would need to change the underlying api call in drone/drone to handle the filtering correctly. Otherwise the end user would see strange behaviour, as we only filter the end result from the api call. EG not displaying anything on the screen. 

Updating the backend api is the way to help enable this, but it opens a can of worms for filtering and we should design that properly. 

I am not going to close the PR but leave it open to see, if others comment on the functionality. We are working on getting a RFC process up and running so we can design features like this in the open, going forward.

julienduchesne:
Hi @tphoney, thanks for reviewing. I did see that there's not much filtering going on in the API (except perhaps here for branch: https://github.com/drone/drone/blob/master/handler/api/repos/builds/list.go). For example, most of the CLI filtering happens in the CLI code directly. I'll be happy to open an RFC when it's possible.

tphoney:
Hey @julienduchesne  we announced the proposals site a few days back https://discourse.drone.io/t/drone-proposals/9415, it lives here https://github.com/drone/proposal 
I am currently working on a proposal for an aws runner.

julienduchesne:
Hey @tphoney, I opened a PR on the Drone API to have filtering by event: https://github.com/drone/drone/pull/3103

PR #357, opened 2021-06-15
(bugfix) render logs if there is an error on build screen

tphoney:


jvrplmlmn:
@tphoney for my own understanding, would this provide additional context in scenarios like

image

?

tphoney:
from what i remember, yes it does.

jvrplmlmn:
Nice! Getting this patch in would be really useful for us 😃

d1wilko:
this bug will be fixed by this PR - https://github.com/drone/drone-ui/pull/365

I'm going to close this PR

PR #358, opened 2021-06-30
Defect/dron 91 trailing slash

eoinmcafee00:
Bug fix: 

We just noticed that if you're calling a Drone project with a trailing slack, like 

https://drone.your.domain.com/githubOrg/Repo/ 

then all the build URLs are formatted with a double // like: 

https://drone.your.domain.com/githubOrg/Repo//12345

Which is leading to a 404 page not found error.

I'm not sure it's a regression from the v1 UI? Maybe the v2 UI router is not sanitizing the URL as it should?


src/index.jsx
```diff
@@ -1,12 +1,15 @@
...
 import ReactDOM from 'react-dom';
+import { BrowserRouter as Router } from 'react-router-dom'
 
 import App from './app';
 import './styles/app.scss';
 
 ReactDOM.render(
   <React.StrictMode>
-    <App />
+    <Router>
```
eoinmcafee00, line 10: why this change? https://flaviocopes.com/react-router-uselocation-usehistory-undefined/

PR #361, opened 2021-07-16
allow ~ in project name

muffl0n:
Bitbucket server allow users to create private repositories. The path looks like this: gitserver.company.com/~muffl0n/drone-test.git

This PR adds ~ as a allowed char in the project name


muffl0n:
See https://regex101.com/r/zjW3KA/1

muffl0n:
Thank you! 😘

PR #372, opened 2021-08-30
fix branch link bug

rohow:
Fix the wrong field reference in the component `PullRequestActivity` to fix the branch link


d1wilko:
hey @rohow good catch, thank you - I added a few other fixes around this - merging now :)

PR #376, opened 2021-09-02
add runner machine to graph step info drawer

d1wilko:


src/components/pages/build/graph-view/step-info-drawer/step-info-drawer.jsx
```diff
@@ -75,4 +79,29 @@ const StepInfoDrawer = (props) => {
   );
 };
 
+StepInfoDrawer.propTypes = {
```
tphoney, line 82: this is a question not a criticism, are these prop types being used for the state / refresh issue ?

d1wilko: they are there in settings.jsx and general.jsx - but they are not actually required / have no material functionality.

they are mainly there so that developers coming behind know what types of props should be passed, throw warnings if the type of props is wrong and also to keep the linter happy :)

tphoney (approved):

Issue #363, opened 2021-07-29
High CPU usage on browser tab during log window changes


joseluisq:
I'm using latest Drone [2.0.4](https://github.com/drone/drone/blob/master/CHANGELOG.md#204) but looks like your DRON-97 issue is not resolved.
When I trigger a pipeline with some Docker steps and one does consonantly a lot of changes to display in log screen then if for example I switch from one previous step to that active step, then I activate the top-right icon to stick the log window during scrolling. It just gets frozen consuming unnecessary amount of CPU. If you compare it with the second tab in the screenshot which is Github Actions (running at the same time than Drone).

![2021-07-29_21-51](https://user-images.githubusercontent.com/1700322/127557171-3e1e788a-8d7e-4923-ac6b-c5a8aac360ff.png)

Here the steps to "try" to reproduce it.
I say "try" because I don't know precisely when it just happens.
Play with steps in order to reach the issue.

- Trigger a pipeline with more than one step. (please uses a pipeline with at least one heavy step which can generate a lot of log lines per second)
- Wait until first finishes
- Switch to the second one (the sticky icon is not activated yet)
- Then scroll top to bottom many times
- Now activate the sticky icon
- Then scroll top to bottom many times again
- At some point you will note that the GUI log changes should get your entire tab frozen with many CPU wasted

Finally, but not less important. The above issue happens to me on a personal testing environment. However I know from other people that uses latest Drone 2.0.4 and they face the same problem.

PR #389, opened 2021-10-17
Remove svg animation

### LINKED ISSUE #315: Animated SVGs causing 50-100% load on 1-2 CPU cores
tl;dr; Opening drone tab with a pending/in progress job makes my laptop CPU temperature increase by 40 degrees within 1.5 minutes

SVG animations are not GPU accelerated and their performance haven't been given much attention by browser developers. 
The tiny animations indicating a pending / in progress job are enough to have 1-2 cores running at 50-100% in both Chrome 78 and Firefox 70.

Those animations are fixed size right now, so an animated gif would look exactly as crisp as svg and negligible amount to page size (2kb maybe?).

https://discourse.drone.io/t/drone-ui-very-high-cpu-usage-because-of-animated-svg-spinner/6110/2

albertyw:
This removes the infinite spin animation on the running, pending blocked, and waiting status SVGs (for reference, the SVGs are at https://github.com/drone/drone-ui/tree/main/src/svg/status).  

This should fix #315 and likely #363.

### DIFF
--- a/src/components/shared/status/status.module.scss
+++ b/src/components/shared/status/status.module.scss
@@ -50,14 +50,6 @@
     background: #09c270;
   }
 
-  &-running,
-  &-pending,
-  &-blocked,
-  &-waiting_on_dependencies {
-    svg {
-      animation: spin 2s linear infinite;
-    }
-  }
   &-pending,
   &-blocked,
   &-waiting_on_dependencies {


bradrydzewski:
thanks for the pull request, however, we are not interested in removing animations from the application.  The issues you referenced are related to lack of gpu acceleration of svg animations, which has since been resolved in chrome 89+ and firefox [1]

> SVG is a great addition to any website, and now those interactions with SVG can be more performant. As of Chromium 89, Chrome will join the likes of Firefox to enable hardware-acceleration by default on SVG animations. What do you, the developer, need to do? Nothing—this will be automatically applied for SVG animations in Chromium 89+.

[1] https://developer.chrome.com/blog/hardware-accelerated-animations/#hardware-accelerated-svg-animations

PR #395, opened 2021-12-13
New filter and sort options for repo list

d1wilko:
- Add ability to sort repos by recency or alphabetical title
- Add ability to filter by organisation
- Fix active toggle logic


abdennour:
If the list of organizations is huge, the UI is hanging.
Think about progressive load of the list.

PR #398, opened 2022-01-05
strip trailing slash from url to fix build link bug

d1wilko:
also general lint fixes


eoinmcafee00 (approved):

bradrydzewski:
could this impact multiple pages?  wondering if this could be solved at the router level instead to trim all trailing slashes?
https://jasonwatmore.com/post/2020/03/23/react-router-remove-trailing-slash-from-urls

d1wilko:
@bradrydzewski yes it can :) I've updated the PR

PR #401, opened 2022-01-10
allow non-admin user to toggle repo protected status

d1wilko:
non-admin user's ability to toggle repo protected status was removed as a a regression between v1 and v2 - this will fix that


tphoney (approved):
looks cool, just moving some lines

PR #432, opened 2023-04-05
Prepopulate the new-build-form with query params

kgwebsites:
Currently manual drone promotions have be done by going into a drone build, clicking the kebab menu and clicking Promote.

Then the user has to fill out all of the fields necessary to promote their build. This is pretty user error prone. 

We do not want to automate this process completely for the time being (which we can currently already do via cli), we still want a user to have to manually trigger this drone promote when they are ready.

This PR will allow us to generate a drone url such as `/:namespace/:name/:build?target=production&commit=1234&parameters=[{"id":"1","key":"IMAGE_TAG","value":"1234"},{"id":"2","key":"ACTION","value":"deploy"}]` . We can then give this generated link to our developers via slack (or anywher else) so they can easily click a link and click Promote without having to manually type out the parameter fields.

### DIFF
--- a/src/components/pages/repo/new-build-form/new-build-form.jsx
+++ b/src/components/pages/repo/new-build-form/new-build-form.jsx
@@ -11,11 +11,17 @@ import css from './new-build-form.module.scss';
 
 const cx = classNames.bind(css);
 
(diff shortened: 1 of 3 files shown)


d1wilko (approved):
LGTM

PR #434, opened 2023-04-17
Function to prefill promote modal w/ query params

kgwebsites:
Previously I added the abillity to prefill a custom build via query params. Duplicating that functionality over to build promotions as well.

### DIFF
--- a/src/pages/build/deployment-form/deployment-form.jsx
+++ b/src/pages/build/deployment-form/deployment-form.jsx
@@ -9,11 +9,11 @@ import css from './deployment-form.module.scss';
 
 const cx = classNames.bind(css);
 
-const DeploymentForm = ({ handleSubmit, handleCancel }) => {
+const DeploymentForm = ({ handleSubmit, handleCancel, action, target, commit, parameters }) => {
   const [state, setState] = useState({
-    action: 'promote',
-    target: '',
-    parameters: [],
+    action,
+    target,
+    parameters,
   });
   const [parameterState, setParameterState] = useState({
     key: '',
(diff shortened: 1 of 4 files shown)


praxist:
related: https://github.com/drone/drone-ui/pull/432

src/pages/build/deployment-form/deployment-form.jsx
```diff
@@ -9,11 +9,11 @@ import css from './deployment-form.module.scss';
 
 const cx = classNames.bind(css);
 
-const DeploymentForm = ({ handleSubmit, handleCancel }) => {
+const DeploymentForm = ({ handleSubmit, handleCancel, action, target, commit, parameters }) => {
```
d1wilko, line 12: Is the `commit` prop ever used?

d1wilko (changes requested):
hey @kgwebsites small question on this one :)

src/pages/build/deployment-form/deployment-form.jsx
kgwebsites: ah good catch. This was copied over from the repo.jsx. I removed it from both, since it is not a used query param in either of the build/promote modals

d1wilko (approved):
LGTM

src/pages/build/deployment-form/deployment-form.jsx
d1wilko: sounds good - thank you :)

tonglil:
<s>
@kgwebsites what's the URL format for filling the query parameters for the `parameters` field? 

It's not the typical `promote?target=target&key=value` and trying json `target=target&parameters={"key":"value"}` does not seem to work either.
</s>

Can't find any documentation on this, would appreciate guidance here!

Nevermind, after iterating I found out to pre-fill the promotion form through URL parameters looks like this (a json array of kv objects:

```
/promote?target=target&parameters=[{"key":"k","value":"v"}]
```

PR #448, opened 2023-11-30
feat: Show the value in the promotion dialog

nenad:
This PR shows the value of the parameters in the prompt when promoting. It is confusing when you promote through prepopulating the config #432, but don't see the values of the parameters.


nenad:
@d1wilko This one is related to #432. Would be great if you can take a look, thanks!

d1wilko (approved):
lgtm

